PR #13, opened 2015-01-27
Fix bug in Layer::getIntersection which could cause an infinite loop

chrissharkey:
This fixes a bug in getIntersection. If you looked at it before, if "continueSearch" is ever set to true, unless it finds another shape on a subsequent loop (not guaranteed) it can end up in an infinite loop.

I believe the solution is to set "continueSearch" to false before or after "spiralSearchDistance += 1". This change helped for me. 

I do notice that even with the fix in place, this loop runs many, many times before finding a result when this antialiased condition is met, it leads to slow down in my app and I'll personally be disabling it.

### DIFF
--- a/src/Layer.js
+++ b/src/Layer.js
@@ -81,6 +81,7 @@
                     }
                     // if no shape, and no antialiased pixel, we should end searching 
                     if (continueSearch) {
+                        continueSearch = false;
                         spiralSearchDistance += 1;
                     } else {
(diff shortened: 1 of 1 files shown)


lavrton:
In this case we will search only with one pixel near pointer. But antialiased pixels might be bigger. Demo: http://jsfiddle.net/krdu22c4/

Issue #12, opened 2015-01-26
Finalize the fork


jovination:
I noticed that on the main repo the forked link has been removed, the reason for this I am unsure about, however it does bring up a couple thoughts.

Because of all the search engine indexing on the retired repo, it leads people to think that is no longer maintained/updated without some sort of link. KineticJS beats many other repo's in my opinion for one simple reason, it does not try to copy flash. I could be wrong on some other libraries but most I have seen have tried to mimic the flash api's for this sort of thing. KineticJS stays JS and not being someone who started out in flash, but javascript its the perfect library for me.

So I can see 2 ways of getting through this.

1. Put more force in the old repo to point to this one.
2. Start fresh.

If the old repo can continue to point to this one, eventually links and such will move to reflect its new home. Eric did a great job here, but if he is to busy to maintain it I would hate for it to fall into obscurity.

As for starting fresh it would be similar to what other repositories have done with forks, rename them. This way any new documentation will not have to worry about being buried in search pages and it will give it new life. Right now many of the tutorials for KineticJS are actually broken, the examples don't work, so unless they are fixed it would be best to use a new name to bring new tutorials to the top.

I intend to help out with KineticJS however i can, I have noticed a few performance issues that I have been working to get around and would love to contribute that if/when i come up with a solution, however without the main repo being actively maintained, and this one being hard to find unless you visited the repo page when the fork link was there it might not get the attention it needs.

Anyway this is only my opinion.

lavrton:
The reason of removing is pretty simple. I have a message from Eric. My fork as not "official opinion". As he said, one day he may create list of available forks, so link will be returned.

All old tutorials I repaired on home page. http://lavrton.github.io/KineticJS/Getting_Started.html

I was thinking about renaming. But it is REALLY huge work for refactoring and editing manuals. I don't want to spend time for this right now. But it might be good idea.

jovination:
I totally understand the renaming lol I have had to do that sometimes and you always miss something and then gotta hunt for it.

Anyway its good to see the reason behind the removing.

Lastly the main reason for me to talk about the docs is that if you google it, the old ones show up and this repo after the link was removed was pretty hard to find. Considering the amount of work you have done with this fork, and the fact I have been unable to find any others I consider this the next official repo but again this is all opinion.

[...]

PR #15, opened 2015-01-27
Renaming codebase to Konva

jovination:
This pull request is a start, thanks to sublime text it was a lot less of a hassle then I originally thought it would be.

All tests passed within my environment.

This goes along issue #12 this should have some review done before its merged since it has so many changes. Although all those changes are not to any functionality.

There is likely still some issues I will fix, mass find/replace most likely has removed some credits from the code referencing KineticJS.


lavrton:
Thanks.

PR #19, opened 2015-01-29
Update old constant names

olimsaidov:


jovination:
Nice catch, that was a pretty big miss on my part.

lavrton:
Nice. Thanks.

Issue #20, opened 2015-01-30
Load handling


jovination:
I don't think there should be any dedicated way to load images or other assets in this library, because so many other stand alone libraries do it so well. However when handling shapes (images mainly) it gets rather strange to constantly make sure that the asset is already loaded before drawing the shape.

So rather then making an area for loading, an area that can tie into other libraries to handle it, but sync the load into konva for loading, so the callbacks are carried over, and the shape will ensure it has loaded beforehand, or at whatever point deemed by the person using it.

I have added something similar into my own library that uses kinetic (soon to be transferred to konva) and it works pretty well to allow assets to be added, without the need to manually ensure the assets have loaded before the draw.

So something like this.

```javascript
var stage = new Konva.Stage(options);
var layer = new Konva.Layer();

var image = new Konva.Image({
   image: url //if it is already an image object it can assume its loaded
});

layer.add(image);
layer.on('ready',function(){
   stage.add(layer);
});
```

This is of course the result without the messing around with the "load" object to handle and connect to the external library but the idea is there.

so here is an example of that

```javascript
var options = {
   loader: new Konva.Loader({
      add: function(url){
         someloader.add(url);
      },
     progress: function(callback){
         someloader.on('asset',callback);
     },
     complete: function(callback){
        someloader.on('complete',callback);
     }
   });
};
```

a very rough example but it should paint the picture, set up the options for the stage and then any assets later will automatically be managed without the need for extra callbacks, also allows the already powerful event system to be more powerful. It would not remove the ability to do it as it is currently, only add an extra option. It could even cache images that are the same so they never have to fire a load event twice and such.

Anyway an idea that I have been wishing this library had as my use of it uses a lot of images. I would like to see if anyone else has an opinion on it before i start trying to do a pull request for it. the above method is far from what the end product will be though, just the original concept as my own implentation uses PreloadJS forcibly.

lavrton:
In this case user may have unexpected results for actions where loaded image is required. For example filtering:

```javascript
var image = new Konva.Image({
   image: url
});
image.cache();
image.filters(filtersList);
layer.add(image);
layer.draw();
```

We can't apply filter before loading. To avoid this we can create special "stack of actions" to apply after loading. But it is too complicated to implement into library.

jovination:
[...]

PR #21, opened 2015-01-31
Add loader functionality

jovination:
Based on the enhancement in #20 

**DO NOT MERGE** *Needs tests, as well as doc blocks*

All current tests pass.

This is more or less the concept of adding a loader to `Konva`, it works by adding a method to `Konva.Node` to allow load handling. By adding it to Node it can be used in any shape or container.

I was unable to create a jsfiddle due to using filters to showcase it, but here is some sample code.

```javascript
var stage = new Konva.Stage({
  container: 'container',
  width: 300,
  height: 300
});

for(var i = 0; i < 5; i++) {
  layer.add(new Konva.Image({
    x: 50 * i,
    y: 50 * i,
    src: 'yoda.jpg',
    width: 106,
    height: 118
  }));
}

layer.load();

layer.on('progress', function(image){
  image.blurRadius(image.index * 5);
  image.cache();
  image.filters([Konva.Filters.Blur]);
});

layer.on('complete', function(){
  stage.add(layer);
});
```

Any node can call `load` and it will send 2 events as it runs `progress` and `complete`. Each progress event passes the individual node, and complete the container. If used on a single node it would send the same object (so in the case of 1 node there is no need for progress).

Also it adds the loader in a similar way to filters, so different kinds could be added in the future, the "default" uses normal image onload, but things like websocket loading is possible (or jquery/preloadJS). The object can be passed as the first parameter in load so `layer.load(Konva.Loader.Default);` there is also a second parameter that handles if the events should be removed after complete, it is false by default.

So what do you think? I am sure there is many kinks to still work out on this before even thinking of adding it.


lavrton:
I don't like this approach. In my opinion `Container` and `Image` (as any other `Konva.Node`) should be designed for drawing, displaying and events detection only. Nothing else. So I don't like adding new `load assets` role. As you say there are a lot of loader libs available.

So I think there is two good ways:
1. Make loader class (or function) that handle loading assets (with no touch any `Konva.Node`). And this class should be VERY tiny and as small and simple as possible.
2. Or create `konva.loader.js` plugin and store it in separate personal (or org) repo. In this case you (and I and all community) can do MUCH more and make it as complex and functional as you want. So you will be able to add new functions in any `Konva` objects.

jovination:
I guess with what I learned here it is possible to do it on the side, on a node i get the src parameter, and on a container i loop the children and get the src elements.

Loaders are made to be plugins more or less, one can easily be crafted with one function to suite the needs of any loader library or method.

I could switch loader into its own namespace, and it is used like so:

```javascript
var loader = new Konva.Loader.Default({
    container: layer,
    removeEvents: true
});
loader.load();
loader.on('progress',function(){
    //you get the idea
});

PR #37, opened 2015-03-16
Added support for nw.js (node-webkit)

sibartlett:
Updated the runtime check, so that Konva does not require node-canvas and jsdom when being hosted in [nw.js](http://nwjs.io/).


lavrton:
Thanks.

PR #40, opened 2015-03-25
Stage toDataUrl bug

luckyadam:
The method layer.toDataURL should have the argument 'config'


lavrton:
I am not sure that this is bug. Can you create any test?

pronebird:
Layers can be of any size so providing the same config as for stage may break things.

lavrton:
I will close the issue for now. Really need a test to see a wrong result.

Issue #43, opened 2015-04-12
Text rendering quality


pronebird:
Hi,

Is there any way to improve text rendering quality?

* top row is not cached
* bottom is cached (only text is cached)

However, text looks fluffy and blurry even without cache, is it possible that pixelRatio causes this?

![text](https://cloud.githubusercontent.com/assets/704044/7105385/1f94533e-e116-11e4-9bf4-7e5cee3e4ea7.png)

**UPDATE**

I found that retina rendering is not applied on laptops which makes picture blurry:
https://github.com/konvajs/konva/blob/4e23aa89e93bc4a08e25c89fc9c805a88a548942/src/Canvas.js#L10

The text is crisp and clear since I removed that check for mobile browser:

![textretina](https://cloud.githubusercontent.com/assets/704044/7105435/a74d7d6c-e118-11e4-9443-95b4aed80603.png)

Do you think we can let desktop browsers to render nice crisp picture? I don't see zooming problems mentioned in comment, the picture seems to zoom naturally.

PR #44, opened 2015-04-12
Enable devicePixelRatio for desktop browsers

pronebird:
This PR is a part of issue #43 and it needs testing because I only checked it on Safari and FF / Mac.

### DIFF
--- a/src/Canvas.js
+++ b/src/Canvas.js
@@ -2,12 +2,7 @@
     // calculate pixel ratio
     var canvas = Konva.Util.createCanvasElement(),
         context = canvas.getContext('2d'),
-        // if using a mobile device, calculate the pixel ratio.  Otherwise, just use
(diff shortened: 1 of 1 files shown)


lavrton:
Nice. You tested and it works, right? I was not able to test it on desktop with retina.

pronebird:
It works fine on my Mac. I think the original zoom issue vanished over past year. I have some animations of falling stars running non-stop and it's stable 60fps with low CPU usage even if I zoom the whole page.

pronebird:
I found that Firefox reports invalid devicePixelRatio if page is zoomed. This is minor but it's worth to notice that nothing has changed since 2013 when this issue was raised last time on bugzilla. The good thing we cache devicePixelRatio so we'll have at least consistent behavior.

Issue #41, opened 2015-04-11
Regression in attrs initialization flow


pronebird:
Hi,

I am migrating from KineticJS to Konva and everything seems fine except some of attributes are missing on my custom subclass. I pass Kinetic.Image (aka Konva.Image) into `config.image` and it does not make it into attrs dictionary.

Looking at [THIS](https://github.com/konvajs/konva/blob/master/src/Node.js#L624) right now and thinking "You gotta be shitting me".

```javascript
// compatibility
var Kinetic = Konva;

// custom class
var Button = function (config) {
  this._initButton(config || {});
};

Button.prototype = {
  _initButton: function(config) {
    // call super constructor
    Kinetic.Group.call(this, config);
    this.className = 'Button';
  }
};

Kinetic.Util.extend(Button, Kinetic.Group);
Kinetic.Factory.addGetterSetter(Button, 'text', '');
Kinetic.Factory.addGetterSetter(Button, 'image');
```

`config` dump:

```javascript
> config
< Object
fontSize: 16
image: Object
text: "Locked"
__proto__: Object
```

`this.attrs` dump:

```javascript
> this.attrs
< Object
fontSize: 16
text: "Locked"
__proto__: Object

```

**Update:**

Changed this
```js
if (key === CHILDREN || config[key] instanceof Konva.Node) {
```

to this:

```js
if (key === CHILDREN) {
```

Problem solved. It is a very strange unimplemented part of Konva.Node.

lavrton:
Test now. Reopen if need.

PR #45, opened 2015-04-12
Fix attrs initialization regression

pronebird:
Don't know who decided that it's a good idea to silently ignore attributes that contain nodes. Cross-reference against KineticJS shows slightly different picture.

https://github.com/ericdrowell/KineticJS/blob/master/src/Node.js#L542

I have a couple of Q regarding that strange piece of code:

1. What was the original intent? 
2. Why was children key ignored instead of presumably being used to add child nodes into newly created node? 
3. Why was type check for `Konva.Node` introduced too?

If none of that serve any purpose then this PR fixes issue #41.

### DIFF
--- a/src/Node.js
+++ b/src/Node.js
@@ -621,7 +621,7 @@
 
             if(config) {
                 for(key in config) {
-                    if (key === CHILDREN || config[key] instanceof Konva.Node) {
+                    if (key === CHILDREN) {
 
                     }
                     else {


lavrton:
Right. Actually I don't remember the reason of adding such code. But tests are failed for request. I will make deep look into code soon.

pronebird:
Apparently circular reference issue that of unit tests is addressing:

https://github.com/konvajs/konva/blob/master/test/unit/Node-test.js#L986

I have to dig it a little bit longer to understand how circular reference can be a problem in javascript. As far as I am concerned setAttrs does not really go down the chain, otherwise it should check for *this* instead of dropping every node passed in attributes.

pronebird:
I do not have a solution for that. I moved nodes that I attach to custom container outside of `attrs`.

lavrton:
Circular reference may give some problems when we want serialize object. Like `node.toJSON()`.
Also for your custom nested node I recommend you to use `Konva.Group`. Look at `Konva.Label` for example. https://github.com/konvajs/konva/blob/master/src/plugins/Label.js http://konvajs.github.io/docs/shapes/Label.html

pronebird:
Thanks.

I used Group under the hood to build a Button and a lot of that code is based on the same principle as Label.

Essentially Button manages text and shape nodes inside and optionally allows to assign image node. I use custom setter because I need to reference image node somehow. Otherwise it's purely a Group node with managed layout.

Circular references can happen also within plain objects, therefore the class check wouldn't be sufficient enough anyway.

I ran tests today and I saw that infinite loop in toJSON. There is no easy way to fix that and the best would be to leave it to developer.

Issue #46, opened 2015-04-12
Cache issues when using FX


pronebird:
Hi,

I am caching the whole group and I experience some pixelRatio issues. It seems that the larger pixel ratio is, the greater is the cached image in the end. It seems the image is like 4 times larger than on 1x ratio. 1x works fine. Same effect on both iPhone and Mac.

```js
if(window.devicePixelRatio) {
  Konva.pixelRatio = window.devicePixelRatio;
}

var W = 800;
var H = 600;

var group = new Konva.Group();
var backgroundImage = new Konva.Image({
  x: 0, y: 0,
  width: W,
  height: H,
  image: image
});
group.add(backgroundImage);

// removed grid drawing for sanity...

group.cache({
    x: 0, y: 0,
    width: W,
    height: H
});
group.filters([ Konva.Filters.Grayscale, Konva.Filters.Brighten ]);
group.brightness( 0.2 );
```

Latest KineticJS:

![screenshot 2015-04-12 19 13 33](https://cloud.githubusercontent.com/assets/704044/7106665/c51b33d0-e149-11e4-84f0-ff8a15089ddb.png)

Konva:

![screenshot 2015-04-12 19 12 06](https://cloud.githubusercontent.com/assets/704044/7106667/d05cf1ac-e149-11e4-830c-81c0974076e0.png)

I barely understand what's going on in cache internals, but it seems if I return `sceneCanvas` instead of `filterContext` then picture has proper size.

https://github.com/konvajs/konva/blob/master/src/Node.js#L335

It can be that filterContext is somehow improperly configured..

pronebird:
Made a jsbin to display the problem:

http://jsbin.com/duwuqakesi/1/edit

You can comment pixelRatio setup to see that default pixelRatio = 1 works fine.

lavrton:
Test, please.

pronebird:
Fixed.

PR #47, opened 2015-04-12
Fix invalid pixelRatio when using filters

pronebird:
I don't know what I did. Can you please check if it makes sense? 

Fixes issue #46


pronebird:
It seems that when we blit canvas elements we have to work in their visual bounds defined by style attribute. I might be wrong in the cause but I can't find any other explanation because sceneCanvas and filterCanvas are identical in their dimensions and pixel ratio.

PR #49, opened 2015-04-13
Fix hit canvas pixel ratio

pronebird:
The code that handles hit context processing does not take into account pixel ratio which results in invalid hit map because `getImageData` returns more pixels than hit context can possibly hold because it is always created with pixelRatio = 1.

We better have smaller hit maps to save some memory but I could not find an easy way to tackle this. 

I've been looking for solution using `drawImage` on top of hit canvas using a replacement fill color but I couldn't find support for that in current Canvas API implementation, that would be the cheapest way, otherwise we have to loop through hundreds of thousands of bytes and replace colors manually. CPU wise it is probably cheaper to have larger hit map than to resample returned data set or use a temporary canvas for that.

https://github.com/konvajs/konva/blob/master/src/Shape.js#L399

mask3

Legend: 

1. red mask is what we have now
2. green mask is a fixed hit map (this PR)

### DIFF
--- a/src/Node.js
+++ b/src/Node.js
@@ -188,7 +188,6 @@
                 height: height
             }),
             cachedHitCanvas = new Konva.HitCanvas({
-                pixelRatio : 1,
                 width: width,
                 height: height
             }),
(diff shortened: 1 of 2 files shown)


pronebird:
I found a way to scale scene canvas using drawImage(). Last commit maintains pixelRatio for hitCanvas at 1x.

pronebird:
Closing in favor of #51

PR #51, opened 2015-04-14
Hit canvas optimization

pronebird:
1. Blit directly to hit canvas, work only with hit canvas buffer. No need to create an ImageData buffer for sceneCanvas. It's cheaper to work with single buffer.

2. Reset Konva.pixelRatio to default undefined value in Stage tests.


pronebird:
Hmm sorry for a PR based on old master, I apparently mixed something.

pronebird:
I'll make a clean PR.

PR #52, opened 2015-04-14
Blit directly to hitCanvas and avoid using temporary canvas

pronebird:
Take 2:

1. Blit directly to hit canvas, work only with hit canvas buffer. No need to create an ImageData buffer for sceneCanvas. It's cheaper to work with single buffer. I think it's pretty expensive to allocate a temp canvas as opposed to using the same hitCanvas.

2. Reset Konva.pixelRatio to default undefined value in Stage tests.

### DIFF
--- a/src/Shape.js
+++ b/src/Shape.js
@@ -386,47 +386,32 @@
             var threshold = alphaThreshold || 0,
                 cachedCanvas = this._cache.canvas,
                 sceneCanvas = this._getCachedSceneCanvas(),
-                sceneContext = sceneCanvas.getContext(),
                 hitCanvas = cachedCanvas.hit,
                 hitContext = hitCanvas.getContext(),
-                width = sceneCanvas.getWidth(),
-                height = sceneCanvas.getHeight(),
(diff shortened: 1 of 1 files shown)


src/Shape.js
```diff
@@ -386,48 +386,33 @@
...
-                        pixelRatio : 1
-                    });
-                    tempCanvas.getContext().drawImage(sceneCanvas._canvas, 0, 0, hitWidth, hitHeight);
-                    sceneImageData = tempCanvas.getContext().getImageData(0, 0, hitWidth, hitHeight);
-                    sceneData = sceneImageData.data;
-                } else {
-                    sceneImageData = sceneContext.getImageData(0, 0, width, height);
-                    sceneData = sceneImageData.data;
-                }
                 hitImageData = hitContext.getImageData(0, 0, hitWidth, hitHeight);
```
lavrton: I guess it is better to clear context after getting image data.

pronebird: It does not make any difference because code below effectively replaces image data. See "else" branch which lowers alpha to 0.

lavrton:
Thanks.

pronebird:
Yeeeehaw! :dancers:

Issue #58, opened 2015-04-15
Stage.toDataURL() invalid output on Retina


pronebird:
Works fine on pixelRatio = 1 though.

![screenshot 2015-04-16 00 11 18](https://cloud.githubusercontent.com/assets/704044/7170308/57c0d3f4-e3cd-11e4-8656-9032aea233f7.png)

```js
var image = document.getElementById('tiger');

Konva.pixelRatio = window.devicePixelRatio; // 1

var stage = new Konva.Stage({
    container: 'konva-container',
    width: 640,
    height: 480
});
var layer = new Konva.Layer();
var tiger = new Konva.Image({
    image: image,

    // assume tiger is in 2x scale
    width: image.width * 0.5,
    height: image.height * 0.5,
    
    draggable: true
});

tiger.cache();
tiger.drawHitFromCache();
layer.drawHit();

layer.add(tiger);
stage.add(layer);
stage.draw();

var snapshotStage = new Konva.Stage({
    container: 'konva-container2',
    width: 640,
    height: 480
});

stage.toImage({
    callback: function (image) {
        var imageNode = new Konva.Image({
            image: image
        })
        var snapshotLayer = new Konva.Layer();
        snapshotLayer.add(imageNode);
        snapshotStage.add(snapshotLayer);
        snapshotStage.draw();
    }
});
```

lavrton:
The result is might be unexpected for first look. But I think it is correct.

After using `toImage` you have high resolution picture. So you need to scale in when you add it to layer.
In the same way as 2x images are added to page with css (`images/box-bg@2x.png` is 400X400):
```css
.box{
        background:url('images/box-bg@2x.png') no-repeat top left;
        background-size: 200px 200px;
}
```

pronebird:
That's what I thought first too. toImage returns the same 640x480. It should be twice larger for hi-res right?

lavrton:
Right. Might be another (not only 2x) size depend on ratio.

pronebird:
`toDataURL` uses SceneCanvas with pixelRatio = 1. I have tested once again on my game and 2x is horribly broken. The issue is with cached elements only. Non cached elements are rendered properly. I'll look into this when I think, preliminary I think we forgot to shrink image and I bet we use the same `drawImage` to paint cached elements.

PR #61, opened 2015-04-16
Fix Stage#toDataURL to produce retina snapshots

pronebird:
`Stage#toDataURL()` creates a canvas with pixelRatio = 1. As you can see if context.canvas comes with pixelRatio = 1 and cacheCanvas has a ratio of 2, then we have twice larger image on output.

Fixes issue discussed in #58.

Screenshot:

screenshot 2015-04-16 17 13 45

The bottom tiger is a bit blurry, that's because it's 1x scale which toDataURL produces. I guess we can tackle this issue in some other PR.

### DIFF
--- a/src/Node.js
+++ b/src/Node.js
@@ -1391,7 +1391,7 @@
                 canvas = new Konva.SceneCanvas({
                     width: config.width || this.getWidth() || (stage ? stage.getWidth() : 0),
                     height: config.height || this.getHeight() || (stage ? stage.getHeight() : 0),
-                    pixelRatio: 1
+                    pixelRatio: config.pixelRatio
                 }),
                 context = canvas.getContext();
 
--- a/src/Node.js
+++ b/src/Node.js
@@ -307,9 +307,9 @@
             );
 
             var cacheCanvas = this._getCachedSceneCanvas();
-            var ratio = context.canvas.pixelRatio;
+            var ratio = cacheCanvas.pixelRatio;
 
-            context.drawImage(cacheCanvas._canvas, 0, 0, cacheCanvas.width / ratio, cacheCanvas.height /ratio);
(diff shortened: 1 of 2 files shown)


pronebird:
Added more commits to address issues rendering images for retina. 

Here is my test for 1x, 2x and 3x:
partyanimal

You can pull my test stage if you wish, it's in here:

https://github.com/pronebird/konva/tree/playground

Run `gulp server` then go to `http://localhost:8080/playground/playground.html`

If you use firefox you can easily open generated data URL in new tab by running:
```js
stage.toDataURL({ pixelRatio: 2, callback: function (i) { window.open(i); } });
```

pronebird:
Oh forgot to mention. Images produced with `pixelRatio > Konva.pixelRatio` when `Konva.pixelRatio = 1` are as not crisp as with `Konva.pixelRatio = 2` but have correct dimensions. Not sure why, maybe I missed something somewhere.

pronebird:
Any ideas regarding tests here? What kind of tests we could implement to verify the behavior?

lavrton:
I made the test. See https://github.com/konvajs/konva/commit/923c466572eda9aa08bfcc56e1becb5a7f749944

pronebird:
:+1:

lavrton:
Now I think it is better to set `config.pixelRatio = 1` default value.

pronebird:
Would you mind to elaborate? I don't see how one would benefit from having pixelRatio = 1 set as default when working on retina canvas.

lavrton:
When I convert stage 400x400 I expect 400x400  image on any device. Right? Otherwise developer MUST think about current device ratio. So "default way" should just work without any extra steps.
So it is better to see blurred (lower pixelRatio) image than extra large image.

pronebird:
Yeah you'd have to use explicit dimensions to make sure image fits or calculate scale. On the other side, retina aware code would have to detect pixelRatio explicitly, would be good then to expose the helper that we have there to calculate native pixelRatio.

PR #63, opened 2015-04-20
ReferenceError: window is not defined

cactis:
```js
ReferenceError: window is not defined
  at /Users/ctslin/Documents/nodes/pboooks-nodejs/node_modules/konva/konva.js:1165:36
  at /Users/ctslin/Documents/nodes/pboooks-nodejs/node_modules/konva/konva.js:1173:11
```


pronebird:
Seems legit. :+1:

lavrton:
thanks.

PR #68, opened 2015-04-21
Fix color alpha rounding bug. (Fixes #67)

### LINKED ISSUE #67: Color animations in tween
I am not able to animate fill color. I am pretty sure you fixed that before...

```js

var stage = new Konva.Stage();
var circle = new Konva.Circle({
   x: 100, y: 100,
   radius: 40,
   fill: 'rgba(0,255,0,0.5)'
});

stage.add(layer);
layer.add(circle);

var tween = new Konva.Tween({
  node: circle,
  duration: 0.5,
  easing: Konva.Easings.EaseInOut,
  fill: 'rgba(0,255,0,0)',
  onFinish: function () {
    circle.cache();
  }
});

tween.play();
```

It seems that transparency is totally ignored.

pronebird:
Unlike color components range [0..255], alpha component uses [0..1] range. I believe the use of `Math.round` is a typo. Parentheses in code left to avoid string concatenation problems.

### DIFF
--- a/src/Tween.js
+++ b/src/Tween.js
@@ -156,7 +156,7 @@
                             Math.round(start.r + diff.r * i) + ',' +
                             Math.round(start.g + diff.g * i) + ',' +
                             Math.round(start.b + diff.b * i) + ',' +
-                            Math.round(start.a + diff.a * i) + ')';
+                            (start.a + diff.a * i) + ')';
                 } else {
                     newVal = start + (diff * i);
                 }
--- a/test/unit/Tween-test.js
+++ b/test/unit/Tween-test.js
@@ -156,17 +156,25 @@ suite('Tween', function() {
         layer.add(circle);
         stage.add(layer);
 
-        var c = Konva.Util.colorToRGBA('green');
+        var duration = 0.1;
+        var c = Konva.Util.colorToRGBA('rgba(0,255,0,0.5)');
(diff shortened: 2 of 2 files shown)


lavrton:
Thanks. Can you create unit test for this?

pronebird:
Hmm how can we test it?

lavrton:
Look at https://github.com/konvajs/konva/blob/master/test/unit/Tween-test.js#L142
So you need new test with something like:
```javascript
        // checking alpha 
        var endFill = 'rgba(100,100,100,0.5)';
        var tween = new Konva.Tween({
            node: circle,
            duration: 0.1,
            fill : endFill,
            onFinish : function() {
                assert.equal(endFill, circle.fill());
                done();
            }
        });
        tween.play();
```
I am trying to cover any bug or new feature with tests. TDD you know.
So as you are doing useful pull request it would be really nice to have tests with them.

pronebird:
I agree but this bug affects intermediate calculations, start and end values do match.

lavrton:
Hmmm. Ok. So I think we need to find a way to make an assert while tweening.

pronebird:
Math.round rounds to lower or upper bound, so in the middle (t = 0.5) it would always produce error, so that's the value of interest here

pronebird:
I've updated tests to verify color interpolation. That covers alpha bug too.

lavrton:
I was thinking to test alpha channel to be in range. But your test looks good to. Thanks.

PR #97, opened 2015-09-15
Config to not export invisible layers

ManuelRauber:
I recently had the problem, that I had to export a stage which has some invisible layers. Within the stage export, the invisible layers are still drawn.

With this pull request I want to add an option to the stage's "toDataURL" function, which will, when set to true, not export invisible layers.
The option is set to false per default to not break existing applications using Konva.


lavrton:
Hmm... I think skipping invisible layer is expected behavior. So I don't think we need special flag.

ManuelRauber:
Okay, you are right. I checked it again. I have to execute stage.draw() again, after setting setVisible(false). I missed that point! :)

PR #102, opened 2015-10-29
added rotation as property since rotationDeg is obsolete

nmehlei:
Only a tiny change but since using rotationDeg results in undesirable warnings, one should rather use rotation and that obviously requires a corresponding property in the typescript typings.

We could additionally remove the obsolete rotationDeg from these typings but since it still works (albeit with warnings), I guess that would be premature.


lavrton:
Thanks!

PR #126, opened 2016-02-01
Fix TypeScript definition

IgorViskov:
Add "getClientRect()" declaration to "Container" class.


lavrton:
Thanks. Can you move it to `Node` class?

lavrton:
Thanks!

PR #133, opened 2016-02-26
container selector

zhouningyi:
You can use  '.main'  &  '#main' to define  container:
besides:  container: 'main' => container: '#main'


lavrton:
Good pull request. Thanks.

PR #137, opened 2016-03-15
make konva run in node 0.12 atleast

hugomrdias:


src/Global.js
```diff
@@ -29,7 +29,7 @@
  */
 
 // runtime check for already included Konva
-(function(){
+(function(global){
```
lavrton: Why do we need `global`?

lavrton: Global detection already included in https://github.com/hugomrdias/konva/blob/node/src/Global.js#L203

hugomrdias: global is undefined inside IIFE

kanzelm3:
+1

kanzelm3:
This fixes an issue I am having as well, please merge this in.

PR #149, opened 2016-05-12
New feature: Clipping content in the circle shape

brusherru:
Added clipType and clipRadius for clipping content of Konva.Container with a circle shape.


lavrton:
Thank you for the pull request!

But I think this will make API more complex. In fact, I want to have only one property `clipFunc` where users can describe any clipping function they need: rect, circle, ellipse and any other shape. And probably deprecate current clipping function of `Container`.

brusherru:
Thanks for fast response!

I understand that this is not the best solution.
I needed this fix for the current project and I think maybe it will be helpful, until there is no better solution.

About `clipFunc`, as I understand it will works like a `sceneFunc` with native context?
I think this is not the best solution for such a powerful framework. As for me it will be better to set another Konva Object as a mask (like this: `myShape.clip( new Konva.Shape(opts) );` ).
Thus beginners can not learn the native tools to draw on the canvas, and the framework will make it easier to change the mask and it is even possible to use animation on the mask.

I understand that is more complex to write it, but this solution is more flexible and really useful, I think :)

brusherru:
Something like this: https://github.com/brusherru/konva/commit/0d0f481c437099268742637f65b6eb4072a50a20
But this is works only with Konva.Shape and maybe not perfect.

I think thats not hard to improve to work with Konva.Containers too and maybe Tweens.

What do you think about it?

lavrton:
Closing in favor of http://konvajs.github.io/docs/clipping/Clipping_Function.html

Issue #150, opened 2016-05-13
Open paths are not filled


e98cuenc:
A bit surprisingly, SVG and HTML5 canvas allow open paths to be filled. Konva's `Path` however doesn't fill open paths. See this example with the same path rendered with SVG and with Konva:

https://jsfiddle.net/0jbs68ut/1/

PR #151, opened 2016-05-14
Support fill on open paths

e98cuenc:
Fix for https://github.com/konvajs/konva/issues/150

### DIFF
--- a/src/shapes/Path.js
+++ b/src/shapes/Path.js
@@ -85,12 +85,7 @@
                 }
             }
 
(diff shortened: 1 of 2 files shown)


lavrton:
Just for reference why `Konva.Path` works in this way: https://github.com/ericdrowell/KineticJS/issues/721

Actually, I am not sure what behaviour is more correct. But I think having the same result as in browser SVG implementation is more expected.

PR #153, opened 2016-05-29
quick fix for typings removing default export

stephenlaughton:
Made it possible to use typings pretty easily with:
```typescript
import * as Konva from "Konva";
```

I still think the typings need work, but this is a quick fix for now.


lavrton:
Thanks.

PR #154, opened 2016-06-06
Clipping Konva.Container with object (linking object as mask)

brusherru:
What do you think about this solution instead of writing a clip function?

### DIFF
--- a/src/Container.js
+++ b/src/Container.js
@@ -366,23 +366,25 @@
             }
             return this;
         },
-        _drawChildren: function(canvas, drawMethod, top, caching, skipBuffer) {
+        _drawChildren: function(canvas, drawMethod, top, caching, skipBuffer, skipClip) {
             var layer = this.getLayer(),
                 context = canvas && canvas.getContext(),
-                clipWidth = this.getClipWidth(),
-                clipHeight = this.getClipHeight(),
-                hasClip = clipWidth && clipHeight,
-                clipX, clipY;
+                clipShape = this.getClipShape(),
+                hasClip = !!(clipShape);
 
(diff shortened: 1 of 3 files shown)


lavrton:
That approach very interesting and looks awesome.

One thing I don't like about it is that using `Konva.Node` instances is VERY heavy. It will have useless listeners. And more important, it will be not destroyed when we destroy original Node. In this case, we will have a memory leak.

brusherru:
@lavrton I agree with you, but if finalize that decision, it is possible to achieve that any Konva object could be a mask, including the container, and tween.

About destroying and memory leaks. I think, when developer creates a Konva.Node and links it to another Konva.Container as mask, developer must monitor that mask destroyed with masked object. Yes, we have a risk to prevent a memory leak, but in the same time developer could use one Konva.Node as mask for many other Konva.Contrainers.
Of course, it is possible to insure against leaks, and add feature that destroys the associated Konva.Node, but then we should make this feature easily disabled. For example, with the second arguments to the function clip().

What do you think?

lavrton:
I am thinking to merge this PR. But I want to make it with 100% backward compatibility. Can you restore `{x,y,width,height}` API and add your `{shape}`? Also I want to add `clipFunc` API.

lavrton:
Closing in favor of http://konvajs.github.io/docs/clipping/Clipping_Function.html
I think `clipFunc` is the best use case.
Simple rectangle (previous implementation) may cover 80% of all use cases. A custom function will cover other 10% of cases.

Clipping with another `Konva.Node` may be still interesting. You can reopen this issue, but please make it 100% compatible with current implementation.

PR #158, opened 2016-07-08
Fix #157 

### LINKED ISSUE #157: toObject serialization
I'm not sure, but looks like:
// line 3430 in kanva.js
// serialize only attributes that are not function, image, DOM, or objects with methods
if (Konva.Util._isFunction(val) || Konva.Util._isElement(val) ||
    (Konva.Util._isObject(val) || Konva.Util._hasMethods(val))) {
    continue;
}

should be:

// serialize only attributes that are not function, image, DOM, or objects with methods
if (Konva.Util._isFunction(val) || Konva.Util._isElement(val) ||
    (Konva.Util._isObject(val) && Konva.Util._hasMethods(val))) {
    continue;
}

|| replaced with &&

ada99:
Stringify custom attributes-objects


lavrton:
👍

PR #165, opened 2016-08-26
Wrap export = in declare module.

yeerkkiller1:
This allows Konva to be used in files that don't import "konva", for the purpose of code like "var x: Konva.Path". This reduces the number of imports that only exist for typings purposes.

### DIFF
--- a/konva.d.ts
+++ b/konva.d.ts
@@ -828,4 +828,6 @@ declare module Konva {
     }
 }
 
-export = Konva;
+declare module "konva" {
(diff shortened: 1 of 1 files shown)


yeerkkiller1:
On second thought, this actually breaks the typescript compilation with "Exported external package typings file 'E:/git/public_web_sites/www.zazzle.com/node_modules/konva/konva.d.ts' is not a module. Please contact the package author to update the package definition.".

PR #175, opened 2016-09-29
Fixed transform apply/reset for clipping

### LINKED ISSUE #136: Bug scale Stage with cached layer and clipped group.
When you try scale Stage with cached layer and clipped group. Clip shifted somehow.
[example](https://jsfiddle.net/Akcium/h8o1ccav/)

iskiselev:
Fix #136. Before _applyTransform ignored top argument, and context.reset() cleared transform that was initially

### DIFF
--- a/src/Container.js
+++ b/src/Container.js
@@ -386,7 +386,7 @@
                   context.rect(clipX, clipY, clipWidth, clipHeight);
                 }
                 context.clip();
-                m = transform.invert().getMatrix();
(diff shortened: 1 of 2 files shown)


lavrton:
@iskiselev hello. Can you make a test that will fail before the patch, and pass after?

iskiselev:
@lavrton , I've added test cases.

lavrton:
@iskiselev Thanks for the Pull Request. I will review and merge asap.

PR #181, opened 2016-10-24
Passive event listeners

raulanatol:
I've been noticing warnings in Chrome:

"Handling of 'wheel' input event was delayed for X ms due to main thread being busy. Consider marking event handler as 'passive' to make the page more responive."

I've used this [doc](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md) to do it.

Sorry for the spaces between functions. Maybe we could use http://editorconfig.org/ to avoid problems like this.

Regards.

### DIFF
--- a/src/Stage.js
+++ b/src/Stage.js
@@ -1,4 +1,4 @@
-(function() {
+(function () {
     'use strict';
     // CONSTANTS
     var STAGE = 'Stage',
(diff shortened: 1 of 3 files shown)


lavrton:
What if a user will need `preventDefault()` in a event handler?

raulanatol:
It's a good question. 
I'm thinking in create a new events name. WHEELPASSIVE or similar.
This project could be help us: https://github.com/zzarcon/default-passive-events but the problem is... how to define passive or no in Konva components?
For example in React project... https://github.com/facebook/react/issues/6436 could be changed to passive for default.

lavrton:
adding `WHEELPASSIVE` will not help because `wheel` event already attached to a Stage container.

raulanatol:
It's true @lavrton.
Any idea?

lavrton:
For now, I see these ways:
1. make `wheel` event passive, but add some kind of notification for a user that he/she can't use `preventDefault()` (probably it is already some kind of notification implemented natively).
2. I guess we have a warning because hit detection is slow. So need to find out why it is slow. From my last experience,  `getImageData()` works VERY slow on scrolling.
3. Just ignore these warnings.  ¯\_(ツ)_/¯

raulanatol:
I could work in the second point keeping the first point like this PR.
Do you think it's ok?

lavrton:
@raulanatol that would be great.

lavrton:
Just want to note, that looks like this issue is fixed in Chrome. Currently, I can't reproduce it and I don't see warning message.

raulanatol:
@lavrton It's true!
I don't see warnings.
I will to close the PR.

Issue #178, opened 2016-10-06
Errors when running tests in node


leahsteinberg:
Hi! I'm having an issue with Konva when I run it in node for tests. Before the tests actually run, when it imports Konva, I get this error:

```
/Users/leahsteinberg/src/project/node_modules/konva/konva.js:284
    Konva.Collection = function() {
    ^
ReferenceError: Konva is not defined
```

This doesn't stop me from running Konva in the browser, I think this has to do with running it in node (for tests). I have tried adding canvas and jsdom, but that doesn't seem to have helped.
Thank you!

lavrton:
Hi. Sorry for late response. Please create a reproducible example.

leahsteinberg:
Hey! I ended up getting around this by only importing konva when it isn't being run in node.

arusantimo:
@lavrton @leahsteinberg  I don't think not-importing konva is a right solution...
I'm working on a server-side rendering with react-konva and this is messing stuff up

PR #185, opened 2016-11-11
global use first

arusantimo:
- Use global first instead of window.
- For NodeJs environment with global.window shimmed using JS-DOM, Konva fails with 'Konva not found' error due to 
    1. Registering Konva on window (which would be global.window)
    2. Trying to find Konva (which would be global.Konva)

Fix for: #178 and


lavrton:
Thanks!

PR #202, opened 2017-02-05
fix ts definition for preventDefault and fill-colorStops

devdoomari:
Files edited: konva.d.ts
added: preventDefault
fixed: fill * colorStops on konva.d.ts  ([string] --> Array<string|number>)


lavrton:
Thanks!

PR #203, opened 2017-02-06
add global alpha value check

ruiann:


ruiann:
if user set the global alpha value, for example
```javascript
new K.Shape({
    sceneFunc: function(context) {
        let bezier = self.getPosition();
        context.beginPath();
        context.moveTo(bezier[0][0], bezier[0][1]);
        context.bezierCurveTo(bezier[1][0], bezier[1][1], bezier[2][0], bezier[2][1], bezier[3][0], bezier[3][1]);
        context.setAttr('strokeStyle', self.options.style.color);
        context.setAttr('lineWidth', 2);
        context.setAttr('globalAlpha', 0.4);
        context.stroke();

        // need to do this manually
        context.setAttr('globalAlpha', 1);
        context.strokeShape(this);
   },
   stroke: 'transparent',
   strokeWidth: 5
});
```
to draw a bezier line, the global Alpha setting will be 0.4 when stroke shape to get the hit context, so the colorKey's value will not be used correctly, hit detection will fail

ruiann:
I will fix the test cases if this PR accepted

lavrton:
I recommend just do not to use `alpha`, `fill`, `stroke()` in `sceneFunc()`:

```javascript
new K.Shape({
   stroke: self.options.style.color,
   strokeWidth: 2,
   opacity: 0.4,
    sceneFunc: function(context) {
        let bezier = self.getPosition();
        context.beginPath();
        context.moveTo(bezier[0][0], bezier[0][1]);
        context.bezierCurveTo(bezier[1][0], bezier[1][1], bezier[2][0], bezier[2][1], bezier[3][0], bezier[3][1]);

        context.strokeShape(this);
   },
});
```

Also, Resetting alpha in this way may decrease performance.

ruiann:
@lavrton what if user want to change the alpha/fill dynamically? This is my situation so I have to manually reset the global alpha value. As you can see for dynamic bezier curve, it's hard to use high level API of Konva, I have to do this work around

lavrton:
Why is it hard? Just update opacity of a node when you need?

Even if it is hard - I think it will be better not write custom `hitFunc` without alpha and fill in its body:
```javascript
new K.Shape({
    sceneFunc: function(context) {
        let bezier = self.getPosition();
        context.beginPath();
        context.moveTo(bezier[0][0], bezier[0][1]);
        context.bezierCurveTo(bezier[1][0], bezier[1][1], bezier[2][0], bezier[2][1], bezier[3][0], bezier[3][1]);
        context.setAttr('strokeStyle', self.options.style.color);
        context.setAttr('lineWidth', 2);
        context.setAttr('globalAlpha', 0.4);
        context.stroke();

        // need to do this manually
        context.setAttr('globalAlpha', 1);
        context.strokeShape(this);
   },
   hitFunc: function(context) {
        let bezier = self.getPosition();
        context.beginPath();
        context.moveTo(bezier[0][0], bezier[0][1]);
        context.bezierCurveTo(bezier[1][0], bezier[1][1], bezier[2][0], bezier[2][1], bezier[3][0], bezier[3][1]);
        context.strokeShape(this);
   },
   stroke: 'transparent',
   strokeWidth: 5
});
```

ruiann:
@lavrton I mean draw a dynamical bezier curve, need to use sceneFunc but not high level API like Konva.Arc and so on, so I need to manually do the alpha value change. If use hitFunc I think it can be solved. But as you can see getPosition is dynamical and will get a random value, I have to do the hit context in sceneFunc. So I think the hit context need to avoid the influence of alpha value, and also I see the use of strokeShape in s_sceneFunc of Rect.

lavrton:
@ruiann even if you will use `sceneFunc` for hit function too, it will be called twice (for scene canvas and for hit canvas). So I think generating random values in `sceneFunc` or `hitFunc` is not a good idea.

ruiann:
@lavrton This is not the main target of this PR. I now know your opinion, you think that when use lower level API, user should keep it work by their own code?

lavrton:
I will close this PR for now. I don't think we should touch `globalAlpha` in hit canvas. Feel free to reopen.

PR #204, opened 2017-02-07
Added support for text strikethrough

kudlajz:
Hi!

Our company is in need of strikethrough in text and we would like to (possibly) use `underline` at the same time, so I've created this PR with strikethrough support.

I've changed the stroke height a bit (for the underline as well) as it was really thin. Now it looks really good, but if you don't like it, feel free to change it or maybe add `textStrikeHeight` property to allow control of it as well.

I have also added test for this use-case, however, I'm not sure where to generate the `trace` string to have something to compare to.


lavrton:
@kudlajz Hello. Very good PR. Isn't it better to add it under `textDecoration()` property (like css works)?
```javascript
node.textDecoration('underline');
node.textDecoration('line-through');
node.textDecoration('underline line-through');
```

kudlajz:
@lavrton Yeah .. I was wondering whether to use array of strings as the parameter to `textDecoration` .. but I can rewrite it this way.

kudlajz:
@lavrton Okay, I have added the strike to textDecoration as you have mentioned. Can you help me with the tests? I am not sure where to generate the content so I can compare to it.

Do I just create the same canvas with the same element as in the test and then copy it there?

lavrton:
Thanks for PR. I added tests, they are very stupid now: it is just copy-paste of current `layer.getContext().getTrace()`. But they ensure that we don't break expected trace when we implement something new.

kudlajz:
@lavrton Great. When are you gonna publish new version?

lavrton:
@kudlajz https://github.com/konvajs/konva/releases/tag/1.4.0

PR #208, opened 2017-02-16
fix clipFunc definitions

devdoomari:
fixed clipFunc typescript definitions


devdoomari:
thanks for merging :+1:

PR #210, opened 2017-02-22
Added dynamically expandable text (x and y) and verticalAlign

kilroy42:
Added one boolean flag and three values to Konva.Text:

- {Number} [config.scaleDynamicX] scale dynamically until either the maximum width is reached or the scaling factor scaleDynamicX is hit. Forces one line text
- {Number} [config.scaleDynamicY] scale dynamically until either the maximum height is reached or the scaling factor scaleDynamicY is hit. Forces one line text
- {Number} [config.scaleDynamicMaxRatio] Limits scaleDynamicX/scaleDynamicY to prevent text thinning or flattening. The default is 4
- {Boolean} [config.verticalAlign] Vertically align the text. Forces one line text.


lavrton:
Hello. First thank you very much for your interest in Konva!

But for currently I have to close the PR, because it is unclear what it is doing and what bugs or new features it resolve.

Please create issue first and describe what you want to archive.

Also we don't need to touch `konva.js` or `konva.min.js` files, because they are generated. All source should be inside `./src`. In case of this PR in `./src/shapes/Text.js`.

kilroy42:
Am 24.03.2017 um 15:23 schrieb Anton Lavrenov:
> Hello. First thank you very much for your interest in Konva!
>
> But for currently I have to close the PR, because it is unclear what it
> is doing and what bugs or new features it resolve.
>
> Please create issue first and describe what you want to archive.
>
> Also we don't need to touch |konva.js| or |konva.min.js| files, because
> they are generated. All source should be inside |./src|. In case of this
> PR in |./src/shapes/Text.js|.

Understood.

I will create a new PR and attach a demo.

Should I only check in src/* or should I assemble everything and check 
everything in once I am done? (src/* + js + min.js)

-- 
Michael

PR #215, opened 2017-03-19
use transform safe offsetX instead of clientX

benbenbenbenbenben:
clientX is not desirable for two reasons.

1. if a css transform is applied, the coordinates may not be correct
2. clientX needs to take the element position into account to calculate a relative position

offsetX/Y avoids both of these problems

### DIFF
--- a/konva.js
+++ b/konva.js
@@ -10393,13 +10393,13 @@
         if (evt.touches.length > 0) {
           var touch = evt.touches[0];
           // get the information for finger #1
-          x = touch.clientX - contentPosition.left;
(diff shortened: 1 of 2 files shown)


lavrton:
Can you create an example where this update make sense?

benbenbenbenbenben:
It makes sense in all situations because it is always easier to use the real values directly. That said, off the top of my head try this:

```<div style='transform:scale(0.76)'>
    <div id='konvaContainer'></div>
</div>```

Your calculated positions will always be wrong.

lavrton:
I agree. Thanks for good PR.

PR #216, opened 2017-03-24
Node.addTo() and Line.addPoint()

mohamed-selim-a:


lavrton:
Thanks for PR, but I don't think it is very useful. We already have `node.moveTo(parent)`. `addPoint()` can be useful, but I want to avoid adding new methods to API, because we already have too many of them, so it is better to do a clean up.

mohamed-selim-a:
node.addTo() is different to node.moveTo(), it's like parent.add(node) but instead of calling it in a separate line, we can add the node to it's container immediately after creating it at the same line.
var rec= new Rect().addTo(group);

PR #235, opened 2017-05-16
Update Arc definition

ChamNouki:
Arc class need to extend Ring in order to be able to access inner and outer radius, and angle method is in Arc class not in Ring.

### DIFF
--- a/konva.d.ts
+++ b/konva.d.ts
@@ -561,16 +561,23 @@ declare module Konva {
         outerRadius(outerRadius: number): Ring;
     }
 
-    interface ArcConfig extends RingConfig {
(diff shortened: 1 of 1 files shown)


lavrton:
You can see in the source that Arc is extended from `Konva.Shape`. We just should add `innerRadius` and `outerRadius` to arc.

ChamNouki:
Ok. So adding properties and methods to ArcConfig and Arc directly extending ShapeConfig / Shape according to sources.
But still removing angle method from Ring which hasn't them.

lavrton:
Looks better now. Thanks!

PR #250, opened 2017-07-24
Fix _mousewheel for firefox (resolve #240)

### LINKED ISSUE #240: Duplicate call this._mousewheel(evt);
https://github.com/konvajs/konva/blob/master/konva.js#L10426
https://github.com/konvajs/konva/blob/master/konva.js#L10438

https://stackoverflow.com/questions/7154967/how-to-detect-scroll-direction

Firefox bug:

default

pavel-pimenov:


lavrton:
Thanks for PR. But what if we will remove `mousewheel` event? Right now I am not sure why we are using it. From https://developer.mozilla.org/en-US/docs/Web/Events/mousewheel it is not standart event.

pavel-pimenov:
I will create separate PR

PR #274, opened 2017-09-12
Added more TypeScript definitions for Konva.Context and Konva.Node.

slig:
Hi,

I added some missing TypeScript definitions. Mostly copied/pasted from `lib.es6.d.ts` with small modifications to match the signature used within Konva.

Before merging, I think it's a good idea to have someone experienced with TS to review it, since this is my first time writing definitions.

Thanks!


lavrton:
Looks good to me! Thanks!

PR #289, opened 2017-11-02
Add scaleDynamicX/Y/MaxRatio, verticalAlign

kilroy42:
I implemented a way to dynamically stretch a Konva.Text to fit a given Konva.Rect since I needed this for at least two of my applications (more to come! I love Konva!)

- scaleDynamicX: scale dynamically until either the maximum width is reached or the scaling factor scaleDynamicX is hit. Forces one line text
- scaleDynamicY: scale dynamically until either the maximum height is reached or the scaling factor scaleDynamicY is hit. Forces one line text
- scaleDynamicMaxRatio: Limits scaleDynamicX/scaleDynamicY to prevent text thinning or flattening. The default is 4

Additionally I added a verticalAlign property to vertically align single line text.

See this example:
https://konva.81818.de/


lavrton:
Hello. Thanks for the Pull Request. We need to think how to apply this changes better.
From my point of view, having `scaleDynamicX` and `scaleX` properties together is very confusing.

kilroy42:
Maybe adding a flag "dontScaleBeyondParent" or "fit/keepInParent" would be a good idea (and then reusing scaleX and scaleY). If this fits better, I would change it this way. scaleDynamicMaxRatio would stay then (?).

kilroy42:
Will close this and reopen as another pull request, I have changed a lot of things, maybe this fits better.

Issue #309, opened 2017-12-20
TypeScript .d.ts class Node does not contain the 'hasName' 'addName' fucntion


iceafish:
when i use typescript write Konva Object, like group, circle. Hit the function 'hasName' 'addName' function not definition. can you fix .d.ts file?

lavrton:
We can. You can do too and send Pull Request. 😉

PR #310, opened 2017-12-20
add addName hasName interface definition

iceafish:
add addName hasName interface definition


iceafish:
fix issue #309

lavrton:
Thanks!

Issue #334, opened 2018-02-07
Add kerning support


EugeneSqr:
Current implementation does not support custom kerning tables. It relies on browser's implementation instead. Having full control over font kerning allows to render texts the same way across all browsers.

Take a look at this [jsfiddle](https://jsfiddle.net/3ct0e6sd/1/). It displays the same text three times (from top to bottom):
- kerning disabled with css (no konvajs)
- kerning enabled with css (no konvajs)
- kerning is always enabled, when browser supports it (with konvajs)

Here is how this jsfiddle looks in different browsers:
1. Chrome, Firefox, Safari. Pay attention to the difference in distance between "A" and "V"
![chrome-kerning-depends-on-css](https://user-images.githubusercontent.com/1623068/35911376-b7591214-0c0a-11e8-99b9-1741f65ed44f.png)
2. IE. The distance is the same no matter what.
![ie11-no-kerning-no-matter-what](https://user-images.githubusercontent.com/1623068/35911494-00dfd3dc-0c0b-11e8-96f0-51d6649c0a4d.png)

lavrton:
Closed by https://github.com/konvajs/konva/pull/337

PR #337, opened 2018-02-08
Added kerning support (issue #334)

EugeneSqr:
Please see the updated [jsfiddle](https://jsfiddle.net/3ct0e6sd/3/).

This time letter spacing should be the same for all browsers (even if they don't have native support for kerning):

chrome

ie11


lavrton:
I think we may need it for both `Konva.Text` and `Konva.TextPath`.
Also, it will be really good to have docs (jsdoc) and unit tests.

@EugeneSqr Do you want to make such updates?

EugeneSqr:
I can handle the docs and unit tests, however I believe we should keep Konva.Text unchanged.

Konva.TextPath calculates each letter position "by hand", therefore it's too slow, when it comes to dragging or moving a lot of text. At this point we can use Konva.Text for "heavy" but "forgiving" scenarios and keep Konva.TextPath for pixel perfect but slow things.

EugeneSqr:
@lavrton it's done

lavrton:
Thanks!

PR #339, opened 2018-02-12
textDecoration typings added

simutaroman:
Hi. Need this textDecoration typings in my project, written with Typescript. So, it could be good idea to add them to main branch.


lavrton:
Thanks!

PR #342, opened 2018-02-16
Add preventDefault option to NodeConfig (type definition) + Add Konva.Text.ellipsis

hbroer:
This config option is important to overwrite the default behaviour. 

Example:
<Rect {...props} preventDefault={false} />


hbroer:
meanwhile i added text ellipsis.
if Konva.Text config is set to wrap='none' and ellipsis=true, then it will add ... to the string.

lavrton:
There must be separate PR.

And I don't get why `ellipses` is useful. Isn't it simpler to cut string manually?

hbroer:
Then i need to figure out how to split it. I will do.

The Konva.Text Element does automatically cut strings if they can't fit into the frame. That is good. But without that ellipsis option u can't define that it should make that visible for the user. Doing that outside of where the cutting happens would make it much more complicated.

CSS has a option for this too: "text-overflow: ellipsis"

lavrton:
Please, make an another Rull Request.

Also `preventDefault` should be part of `ShapeConfig` (not `Node`).

BTW, thanks for the contribution.

hbroer:
This can be canceled. I splitted it up.

PR #354, opened 2018-03-10
Add dynamic text resizing, add vertical alignment, fix JUSTIFY bugs

kilroy42:
Second pull request with modified names and slightly modified behaviour and cleaner implementation.

Demo is here:
https://konva.81818.de/index.html

- Allows text to be dynamically expanded or shrinked inside a given container:
   * @param {Number} [config.resizeToFitX] scale dynamically until either the maximum width is reached or the scaling factor resizeToFitMaxRatio is hit
   * @param {Number} [config.resizeToFitY] scale dynamically until either the maximum height is reached or the scaling factor resizeToFitMaxRatio is hit
   * @param {Number} [config.resizeToFitMaxRatio] Limits resizeToFitX/resizeToFitY to prevent text thinning or flattening. The default is 4

- Allows text to be vertically centered or aligned at the bottom:
   * @param {String} [config.verticalAlign] can be top, middle or bottom

- Fixes JUSTIFY bug on strike-through/underline (lines not long enough)
- Fixes JUSTIFY bug on padding (will leave bounding box)
- Can be seen here (box 8 show both errors): https://konva.81818.de/index2.html


igorcouto:
Is there a milestone to integrate this pr?

lavrton:
I used fixes from that Pull Request to fix some justify text bugs.
I don't think we should have `resizeTo...` properties in Konva core. It makes API too complex.

Probably vertical-align is still interesting feature, but it is too simple to align text vertically manually.

kilroy42:
Sad to hear because I think this is one of the most useful Text properties since its near impossible to know the size of a text before you actually draw it. And calculating this afterwards (and after each refresh/resize) just makes it ugly and look like gibberish JavaScript again. I loved the simplicity of Konva and just wanted to contribute to make it better...

I will then implement a fork of the Text class and publish an ResizableText soon - if thats ok with you.

lavrton:
@kilroy42 yes. A calculation of scale can be ugly. But not so much, that code is very simple:

```js
function adjustScale(string) {
   textNode.text(string);
   const scale = requiredWidth / textNode.getTextWidth();
   textNode.scaleX(scale);
}
```

Konva core already has too many methods and properties. Adding new API will make it bloated. So I am thinking how to reduse and simplify API. So we have to think twice when we want to add a new feature.

kilroy42:
This was my first thought, too. But it won't work with anything like: padding, centering, elippsis, wrapping... essentially Konva.Text will become super dumb then, which is sad :-(

I will fork the Text class as another class then and try to integrate changes as they come, but I still think you should think about this again.

I know a lot of people who want these functions and it is no "bloat" as you say - it is basic functionality to draw text on a canvas whose size you don't know (or which you zoom). Please think about it again - How do you render text when you don't know the size of the final object beforehand?

PR #369, opened 2018-03-22
add option to pass fn as a selector in find method instead of a string.

seethroughtrees:
While using the library, I found myself wanting to find by more than just selectors, so I added a .findWhere() method.

This allows you to pass a function into a find method instead of a selector.
I know we could modify the .find method, but it felt like too many changes to be worth it.

Please let me know if this is worth adding, and if any changes need to be made.

### DIFF
--- a/konva.d.ts
+++ b/konva.d.ts
@@ -324,6 +324,7 @@ declare namespace Konva {
     clipFunc(ctx: CanvasRenderingContext2D | undefined | null): void;
     destroyChildren(): void;
     find(selector?: string): Collection;
+    findWhere(fn: (Node) => boolean): Collection;
     findOne<T extends Node>(selector: string): T;
     getAllIntersections(pos: Vector2d): Shape[];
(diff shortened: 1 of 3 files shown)


lavrton:
Yeah. I think it is a useful method. But I think it will be better to overload `find` method. So it can work like this:

```js
group.find('Shape');
group.find(function(child) {
   // child here can be a direct child or child of a child
});

seethroughtrees:
sure.  I can do that instead.

Just to confirm, overloading `find` will require some changes inside the method as its pretty much written to handle only strings.  We'll have to modify `isValidSelector` (which is used by `_isMatch`) and clean up some of the logic depending on the argument type.  

that okay to modify?

seethroughtrees:
per your initial request, I updated the PR to overload the find method.  It should now take string selectors, or a function filter.

lavrton:
Looks good to me. Thanks!

Issue #225, opened 2017-04-18
Arrows with dotted lines dot arrowhead outline too


kingllama:
End result is really weird looking. I couldn't determine if there was a way around this.

<img width="138" alt="dotted_arrow" src="https://cloud.githubusercontent.com/assets/8138186/25142139/c567aeca-241a-11e7-8eae-c6de7583d908.PNG">

http://jsbin.com/yadiwisovi/1/edit?html,js,output

lavrton:
Fixes with #373

PR #373, opened 2018-03-25
DO NOT MERGE: remove dash from arrow-head

seethroughtrees:
When rendering an Arrow that has a dashed line.  Current behavior is, the arrow head is wrapped in a surrounding dashed outline.  Which provides for an unpredictive rendering.

Also addressed in: https://github.com/konvajs/konva/issues/225

The current fix below removes the dash so a user can still apply a stroke to the arrowhead if desired.  But changes the saves and reapplies the dash attribute after the arrow is rendered.

### DIFF
--- a/src/shapes/Arrow.js
+++ b/src/shapes/Arrow.js
@@ -66,7 +66,21 @@
         ctx.closePath();
         ctx.restore();
       }
+
+      this._fillArrowHead(ctx);
+    },
+
+    // removing dash from arrow-head only, but putting it back to the
+    // intial value after render
+    _fillArrowHead: function(ctx) {
+      var isDashEnabled = this.dashEnabled();
+      if (isDashEnabled) {
+        this.dashEnabled(false);
+      }
+
       ctx.fillStrokeShape(this);
+
+      this.dashEnabled(isDashEnabled);
     }
   };
 


seethroughtrees:
no need for review yet.  thought this worked as expected, but it does not.

lavrton:
The Pull Request is very useful. I want to fix it for a long time but did done it yet.

I am thinking to split `sceneFunc` of `Konva.Line` somehow, so it will be simple to reuse it in `Konva.Arrow`. Probably it will help to fix this issue too.

src/shapes/Arrow.js
```diff
@@ -66,7 +66,21 @@
...
+
+      this._fillArrowHead(ctx);
+    },
+
+    // removing dash from arrow-head only, but putting it back to the
+    // intial value after render
+    _fillArrowHead: function(ctx) {
+      var isDashEnabled = this.dashEnabled();
+      if (isDashEnabled) {
+        this.dashEnabled(false);
```
lavrton: I think it is better to avoid such changing because it is triggering attribute changing events.

lavrton: Probably direct changing of `attrs.dashEnabled` will be better. But I am not sure yet.

seethroughtrees:
i agree about splitting the function for Line and Arrow.  The stroke is being added twice because of that function, which i think was part of the confusion for me.  i'll see if i can come up with something.

lavrton:
Sorry, I just merged it...

I changed the fix a bit. Works ok now. Probably we can apply some optimizations for `stroke` function. But I didn't find a good way yet.

seethroughtrees:
no problem, sorry for the delay on this.  thanks for fixing it up!

PR #376, opened 2018-03-27
update npm scripts

seethroughtrees:
this is just to slightly improve the dev experience a bit.  

- some users would prefer not to install `gulp` globally.  so pointing the path to the local versions.  
- also adding `prettier` config lets IDE's match your cli options config when editing.

*the collapsing arrays were done by prettier and are not intentional changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,13 +2,7 @@
   "name": "konva",
   "version": "2.0.2",
   "author": "Anton Lavrenov",
-  "files": [
-    "README.md",
-    "konva.js",
-    "konva.min.js",
-    "src",
-    "konva.d.ts"
-  ],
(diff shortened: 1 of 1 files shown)


lavrton:
`prettier` and `gulp` are in `devDependencies`, so they will be installed locally.

When you run `npm start`, `npm` will go into local dependencies first. So

```
"start": "gulp",
```
and 
```
"start": "./node_modules/.bin/gulp",
```
Will work exactly the same. So I don't think that change is useful.

lavrton:
But `prettier` config can be useful.

seethroughtrees:
whoa! i didn't realize npm did that behind the scenes.  ok.  let me remove.  1 minute.

seethroughtrees:
done.

PR #377, opened 2018-03-30
add color balance filter

hakoiko:
add color balance filter.
node.redBalance(1.1).greenBalance(0.8).blueBalance(1)
the param value could be 0 ~ Infinity


lavrton:
Closing the PR for now. Thanks a lot for your suggestion, but I think currently `Konva` already has too many filters and bloated API for managing props. Currently, I am thinking about simpler/lighter solutions. 

https://github.com/konvajs/konva/issues/380

Issue #382, opened 2018-04-17
Consider adding crossOrigin = "anonymous" to Image.fromURL to prevent tainted canvases


benjaminjackman:
Hi, It might be advisable to enable crossOrigin loading to the fromURL function in Konva.Image

https://github.com/konvajs/konva/blob/6d909f833f69f909c1bb3afab714942650eed709/src/shapes/Image.js#L234-L243

This brief article lays out why, but long story short having an additional line under
https://blog.codepen.io/2013/10/08/cross-domain-images-tainted-canvas/

```javascript
img.src = url;
``` 
of 
```javascript
img.crossOrigin="Anonymous";
```
could help prevent the canvas from getting tainted

( one more article https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image )

humblecoder:
For some reason I'm still receiving the CORS error.  I also tried adjusting `crossOrigin` to `null` after the fact, although it's probably too late within the callback.  Please advise.

lavrton:
`crossOrigin` do not prevent all CORS error. You may have to fix it from the server side.

humblecoder:
@lavrton apologies, you're correct.  Although I had CORS middleware for Laravel, it was a public resource, so it was only being touched by Nginx.  I had to add an exception in Nginx.  Thanks.

hiratariq:
I was working on a project the other day when I kept getting CORS error while using Konva.Image.fromURL but when went back to image.onload method it working. For some reason having crossOrigin set to '', null or 'anonymous' was the issue. I needed to set nothing to get it to work. So I think if there is a higher level Konva.Image.crossOrigin kind of a thing that user can set once in their project depending on their needs that would be better.

PhilCliv:
Going back to image.onload method made it work for me! Thanks hiratariq..

PR #383, opened 2018-04-17
Set img.crossOrigin to Anonymous in Konva.Image.fromURL  to prevent tainted canvas

benjaminjackman:
See: https://github.com/konvajs/konva/issues/382

N.B. There is also: img.crossOrigin = "Use-Credentials"; (though i figure the anonymous case is the most common).
If that is a concern, perhaps this could be add as an additional optional parameter on fromURL called `crossOrigin` that defaults to "Anonymous". If it's set to null then the field won't be set at all, and if needed the client could pass in "Use-Credentials" 

Additional discussion [here](https://stackoverflow.com/a/23123261/121477) about why `img.crossOrigin` is assigned above `img.src`

### DIFF
--- a/src/shapes/Image.js
+++ b/src/shapes/Image.js
@@ -239,7 +239,7 @@
       });
       callback(image);
     };
-    img.crossOrigin = "Anonymous";
+    img.crossOrigin = 'Anonymous';
     img.src = url;
   };
 })();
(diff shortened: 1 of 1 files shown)


lavrton:
Probably we don't need to use `Use-Credentials`. Because users always can create `window.Image` instance manually and apply all required attributes.

PR #384, opened 2018-04-17
Fixed type definitions for find() and findOne()

troeggla:
With the option `noImplicitAny` enabled in `tsconfig.json`, compilation of a project using Konva would fail because the functions `find()` and `findOne()` implicitly had an `any` type. I added proper type definitions for them. Or rather properly added variable names to the function signature, as it seems the original author just forgot to name the parameters.


lavrton:
Thanks.

PR #410, opened 2018-07-02
Fix Transformer to work with scaling

allada:
Transformer currently does not work properly with scaling. This patch makes it use the passed in scale amount and scale all created nodes by that scale.

This problem was caused because Transformer does not know the pixel scale ratio at the time the nodes are created.

This also fixes when an inverted scale is used to show the correct cursor. (negative scales are often used to draw right-to-left or bottom-to-top instead of default top-left-bottom-right).


lavrton:
The Pull Request looks very interesting. But currently, it works weird.

kapture 2018-07-03 at 12 13 18

allada:
oops, I did a last minute change to make it more testable but only tested it on my personal code (which doesn't use shape scale at all).

Please take another look :-)

PR #418, opened 2018-07-20
Support offsetX and offsetY when transformsEnabled is set to 'position'

VladimirTechMan:
Supporting both `x`, `y` and `offsetX`, `offsetY` under the **'position'** option
of the `transformsEnabled` node-property makes it more useful for real apps.

Also, no real benefit in making the auxiliary variable `transformEnabled`
a part of the function closure, thus moving it to the local function itself.

### DIFF
--- a/src/Node.js
+++ b/src/Node.js
@@ -1427,17 +1427,16 @@
       }
     },
     _getAbsoluteTransform: function(top) {
-      var at = new Konva.Transform(),
-        transformsEnabled;
+      var at = new Konva.Transform();
 
       // start with stage and traverse downwards to self
       this._eachAncestorReverse(function(node) {
-        transformsEnabled = node.transformsEnabled();
+        var transformsEnabled = node.transformsEnabled();
 
         if (transformsEnabled === 'all') {
           at.multiply(node.getTransform());
         } else if (transformsEnabled === 'position') {
-          at.translate(node.x(), node.y());
+          at.translate(node.getX() - node.getOffsetX(), node.getY() - node.getOffsetY());
         }
       }, top);
       return at;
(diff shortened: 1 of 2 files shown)


lavrton:
We need unit tests for that changes. I am not sure that adding offset to absolute position will lead to the correct behaviour.

VladimirTechMan:
Sure, no problem. I have added a unit test checking that `getAbsoluteTransform()` produces the same results both with `transformsEnabled` set to **'all'** and with `transformsEnabled` set to **'position'** when the transforms are limited to position and offset only.

In addition to that, to give you a little more confidence here: If the transforms of a node are limited to position and offset and `transformsEnabled` is set to **'all'**, then calling `node.getTransform()` inside `getAbsoluteTransform()` boils down to two consecutive calls:
```
    m.translate(x, y);
    m.translate(-1 * offsetX, -1 * offsetY);
```

Now, inlining those two calls of the method, we will get:
```
    this.m[4] += this.m[0] * x + this.m[2] * y;
    this.m[5] += this.m[1] * x + this.m[3] * y;
    this.m[4] += - this.m[0] * offsetX - this.m[2] * offsetY;
    this.m[5] += - this.m[1] * offsetX - this.m[3] * offsetY;
```

It is easy to check that the above is mathematically equivalent to:
```
    this.m[4] += this.m[0] * (x - offsetX) + this.m[2] * (y - offsetY);
    this.m[5] += this.m[1] * (x - offsetX) + this.m[3] * (y - offsetY);
```

Which, in turn, is what the following method does internally:
`    at.translate(node.getX() - node.getOffsetX(), node.getY() - node.getOffsetY());`

I hope that the above explanation helps.

lavrton:
Thanks for the contributions!

PR #419, opened 2018-07-24
Removing unused value from return statement in _eachAncestorReverse()

VladimirTechMan:
A minor cleanup. When method `_eachAncestorReverse()` does not have
the top node argument, it returns "nothing" (`undefined`). Thus, the
`true` value returned when it has the top argument – and the current
Konva code never checks or uses that returned value – seems to be
just a side effect of copy-pasting some code in the past, or alike.

Unless the function was planned to return a Boolean value for a good
reason (in which case we'd need to add a return value to its end too),
let's remove the unused value from that return.

### DIFF
--- a/src/Node.js
+++ b/src/Node.js
@@ -1052,7 +1052,7 @@
       // func with this because it will be the only node
       if (top && top._id === this._id) {
         func(this);
-        return true;
+        return;
       }
 
       family.unshift(this);


VladimirTechMan:
Well. So far, it was not that random, actually. ☺️ I am missing one feature in Konva that is important to my project. Aiming to try implementing it and then proposing to it you. While prototyping and checking how things work for me in the browsers, I do have to look under the hood time to time. Then, I start seeing things that I would fix or improve. A usual story – for me and many other devs around – on open source projects, I guess. 😉

VladimirTechMan:
Having said that. I may eventually do some random audits too. 😀

PR #422, opened 2018-07-28
Only update the z-indices that do change after reordering nodes

VladimirTechMan:
Right now, when updating the z-index of a node in a container,
the indices of all its siblings are refreshed. However, if the
node changes its z-index from _z1_ to _z2_, we only need to refresh
the z-indices of the node's parent's children in the closed
range [ _min(z1, z2)_, _max(z1, z2)_ ]. For groups and layers with
many children that will be more optimal in many cases.

In addition to the above change, this patch:
* forces the parameter of `Node#setZIndex()` to be in the valid
  range.
* adds the (now absent) unit tests for the methods of `Node` that
  change z-indices of the parent's children.

### DIFF
--- a/src/Node.js
+++ b/src/Node.js
@@ -1052,7 +1052,7 @@
       // func with this because it will be the only node
       if (top && top._id === this._id) {
         func(this);
-        return true;
+        return;
       }
 
       family.unshift(this);


lavrton:
Hello. Thanks for the Pull Request. The tests are very good. But I don't think `_setChildrenIndicesBetween` is useful. Because it makes the code more complex. And it doesn't help with the performance. Because reordering is a very chip operation and it doesn't happen frequently.

VladimirTechMan:
@lavrton Thanks for the feedback, Anton. I did consider the added code complexity too. I agree that updating indices, in its current way, is already quite cheap and fast in the modern browsers, for small to medium size collections of children in containers. In my case, I would need to consider dealing with bigger ones for a specific case (combined with a few other existing techniques, including partial scene-draw optimizations for scrolling larger canvases). There, it would pay off, I think. (I also thought the added code complexity was not that big, honestly.)

In any case, it is your decision as the project owner. I do appreciate your response in any case. (Feel free to use the test cases alone if desirable.)

lavrton:
I merged your tests.

PR #425, opened 2018-08-03
Anchor Box Size for Transformer

adam-t-b:
Switches out 
`  var BASE_BOX_WIDTH = 10;
  var BASE_BOX_HEIGHT = 10;`

for [config.anchorBoxSize] getter/setter to increase anchor boxes on mobile if wanted (I need this!)

### DIFF
--- a/src/shapes/Transformer.js
+++ b/src/shapes/Transformer.js
@@ -1,9 +1,6 @@
 (function(Konva) {
   'use strict';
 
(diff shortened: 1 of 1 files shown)


lavrton:
Good PR. Can you rename `anchorBoxSize` -> `anchorSize` it case we will want another shape for achors?

adam-t-b:
Updated with the suggested naming.

PR #430, opened 2018-08-05
[shapes/Transformer] added customAnchorFunc to create customized anchors

ajdalshov:
This makes it possible to use custom anchors for Transformer.

Usage:

```javascript
const transformer = new Konva.Transformer({
    customAnchorFunc() {
        return new Konva.Circle({
            stroke: "rgb(138, 138, 138)",
            fill: "rgb(49, 49, 49)",
            strokeWidth: 2,
            radius: 8,
            offsetX: 0,
            offsetY: 0
        });
    }
});
```


ygorlf:
@lavrton, "cornerRadius" support would be awesome, currentlty I use a custom resize algorithm because design restrictions on the handler, for my use case I have a circle with a resize icon inside, I think it's possible to hack away and add a image, but didn't figure that out yet. Looking forward to see more flexibility on transformer, keep it up the good work, awesome framework

PR #435, opened 2018-08-19
get point at path

andrewDyakin:
Get point at length of the path


lavrton:
Hmm. Why may we need this method?

ada99:
I have a link (arrow) between two figures. Link represents a complex path.  And link can have a text (e.g. "Yes", "No"). This text by default is placed at the center of the path and its placement can be changed by user. So I use this methods to do it.

PR #448, opened 2018-09-07
canvas-node, add custom font

zhfish:
# canvas-node，add custom font
### step 1
```
in konva-node.js
Konva.setFont(__dirname, '../font/pingfang.ttf', 'pingfang')
```
### step 2
```
in use
  const titleText = new Konva.Text({
    x: 0,
    width: 750,
    y: 0,
    text: '居中居中居中居中',
    fontFamily: 'pingfang',  // fontFamily = setFont's name
    fontSize: 34,
    fill: '#FC4F88',
    align: 'center',
  })
  layer.add(titleText)
```


lavrton:
The Pull Request is very interesting. I am thinking, can we somehow not to touch Konva core? Like new methods in `Utils`. They are related to `konva-node` only. So probably we can do all work in `konva-node/index.js`. No?

zhfish:
A critical point
it need addFont before draw
1. Konva.Text
i have no idea in a short time  
it relationship text width  
for example “align: center”
2. toCanvas()
it can rewrite in konva-node

lavrton:
@zhfish what do you mean by "but the workload is too heavy"? Sooner or later we have to update to node-canvas 2.0.

zhfish:
in node-canvas 2.0, we need change more in Konva core
i think i can try it

lavrton:
I am closing the issue for now, because it is hard to merge it into version 3. But it will be still good to have updated on it.

PR #466, opened 2018-10-05
rotationSnaps

tole42:
the rotationSnaps feature seem to broken. I tried to fix it.. Only tested with chrome 69


lavrton:
Hello. You misunderstand `rotationSnaps` feature. It doesn't limit the rotation to the provided list of values. It just allows setting rotation to values from the list by making shape "sticky" near that points.

lavrton:
https://konvajs.github.io/docs/select_and_transform/Rotation_Snaps.html

PR #514, opened 2018-12-12
Update konva.d.ts for getRGB

cyrilmesvayn:
getRGB returns an object with RGB values, not a string. 
I've set up a type RGB for typescript. 

Feel free to edit to your heart's content. (I've accidentally added whitespace through GitHub's editor on line 14. My bad)


lavrton:
Thanks!

PR #527, opened 2019-01-03
TypeScript definitions: "setAttrs" must accept same arg than constructor

Ivanca:
Any class that extends from Node must take the same argument on setAttrs as its constructor.

TODO: Check for classes missing on this PR, which only fixes classes that extend from Shape


lavrton:
Thanks for the PR!

Currently `konva.d.ts` is manually created. I hope to have it automatically generated from the source. I am transitioning to typescript source. Any help is very welcome here!

PR #529, opened 2019-01-06
When setting new text on Konva.Text, treat undefined width or height as 'auto'

VladimirTechMan:
Similar to how it is handled in `getWidth()` and `getHeight()` of `Konva.Text`, let `_setTextData()` treat undefined values of the '**width**' and '**height**' attributes as if they are set to '**_auto_**', not as if they are fixed.


lavrton:
Yeah, that makes total sense now! Such fixes are useful and I hope to fix many such issues while migrating to the strict typescript.

PR #531, opened 2019-01-08
Do not recalculate additionalWidth on every line of multiline text

VladimirTechMan:
This is just a minor improvement (optimization) to `_setTextData()`:
Flag `shouldAddEllipsis` is not changed inside the loop, thus no need
to recalculate the value of `additionalWidth` for every line of the
original text.


lavrton:
That is very useful. Thanks a lot!

PR #541, opened 2019-01-15
When rendering multi-line texts, check text decoration flags once

VladimirTechMan:
A small improvement to avoid re-testing the presence of **underline**
and **line-through** attributes in the `textDecoration` property on each
line of the text being rendered.


lavrton:
@VladimirTechMan  Thanks a lot of your work and support! 👍

PR #564, opened 2019-01-31
Konva mouse events should accept stages as targets

jktzes:
stages can also be targets of konva mouse events


lavrton:
Thanks a lot for your Pull Request. But I will reject it because we are in transition to typescript for the whole code base. So type definitions will be automatically generated from the source.

I am happy to see pull request for better types in the codebase. I am going to turn on strict mode soon, so types will be better.

jktzes:
Sure! It's good to hear that.

PR #565, opened 2019-02-02
Fix for the performance regression in the updated batchDraw()

### LINKED ISSUE #562: Huge perfomance drop with Konva after the recent "drag&drop refactor"
Hi @lavrton. I just wanted to give you heads-up on the very notable performance regression that happened after commit a0b2f027ba7d624bab0ab8269a0987e8b4ceed36 "drag&drop refactor, update docs". I was not able to investigate it myself yet – just happened to be tied up at the beginning of this week; but I will get to it at some point. Yet, I checked and confirmed that rolling back to the commit one before the above commit immediately resolves the issue for me, while setting the repository code to the referenced commit ID (or beyond it) brings the issue back.

In general: My own app code does lots of updates to the scene composition in near real-time, based on the incoming data. It uses `batchDraw()` for all (re)draws. It also uses a technique somewhat similar to the one used in one of well-known Konva demos: Demo 4 given at https://konvajs.github.io/docs/sandbox/Canvas_Scrolling.html.

Prior to the above commit, scrolling the scene in my app (even while the app was intensively updating the scene graphics in parallel, based on incoming data) was going at near 58-60 fps, which is petty good. After the above commit, the same app can only go at about 0.8-1.5 frames per second, at best, in the exact same conditions. Essentially, it "freezes".

Before I can look at that myself, a quick question / request to you: After the above commit, does Demo 4 of "How to display and scroll a very big html5 canvas?" still work flawlessly in your own tests? (If you did not try it yet, would you mind trying it sometime and sharing your results?)

VladimirTechMan:
The code was checking the "waiting" flags, but they were not set.
This patch corrects that part and makes a few other tweaks to the
updated logic on top of `requestAnimationFrame()`.

Fixes #562.

### DIFF
--- a/src/BaseLayer.ts
+++ b/src/BaseLayer.ts
@@ -249,13 +249,13 @@ export abstract class BaseLayer extends Container {
    * @return {Konva.Layer} this
    */
   batchDraw() {
-    if (this._waitingForDraw) {
-      return;
+    if (!this._waitingForDraw) {
+      this._waitingForDraw = true;
+      Util.requestAnimFrame(() => {
+        this.draw();
+        this._waitingForDraw = false;
+      });
     }
-    Util.requestAnimFrame(() => {
-      this._waitingForDraw = false;
-      this.draw();
-    });
     return this;
   }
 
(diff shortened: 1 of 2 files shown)


lavrton:
Great! Thanks a lot!

PR #567, opened 2019-02-03
When removing a child, set key "parent" to null, don't delete it

VladimirTechMan:
Structural changes to objects are relatively more expensive and
harder to optimize for modern JavaScript compilers, compared to
just changing object's key values. The **`parent`** property of nodes
is already set to `null` now, as part constructing node objects.
Setting it back to `null` on removal / destroy of node objects is
more consistent and a tiny little bit more efficient.

_P.S. Going forward, time permitting, I may touch a few other places
in Konva that rely on `delete` without much need in using it._

### DIFF
--- a/src/Container.ts
+++ b/src/Container.ts
@@ -66,7 +66,7 @@ export abstract class Container extends Node {
     for (var i = 0; i < children.length; i++) {
       child = children[i];
       // reset parent to prevent many _setChildrenIndices calls
-      delete child.parent;
+      child.parent = null;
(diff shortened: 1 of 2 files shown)


lavrton:
Good change. Thanks.

Also, my goal for the `v3` of Konva is to make sure that all properties are initialized in the constructor of each class. So it will not produce new [hidden classes](https://draft.li/blog/2016/12/22/javascript-engines-hidden-classes/) later .

PR #594, opened 2019-03-02
Removing now-unused method Stage#_onContent

VladimirTechMan:
...Unless there are some specific plans to return to using it, in the future. It has been fully replaced by `Stage#_bindContentEvents` as far as I can see.


lavrton:
Thanks!

PR #614, opened 2019-03-21
Fixed when the font is smaller than 12px and align is center, text ca…

AlfredMou:
### Version
3.2.2
### Demo
 [http://plnkr.co/edit/StWNJ3ZBf7F7aamtSRCx?p=preview](http://plnkr.co/edit/StWNJ3ZBf7F7aamtSRCx?p=preview)
### Steps to reproduce
There is a problem in the chrome browser.  When the Konva.Text font is less than 12px, Unable to center alignment. Because the minimum width of the canva measureText can only be taken to 12px. 
My chrome version is 72.0.3626.121 .Firefox and safari are normal, and some chrome performance is normal. Depending on whether the browser can render fonts smaller than 12px

### DIFF
--- a/src/shapes/Text.ts
+++ b/src/shapes/Text.ts
@@ -93,6 +93,17 @@ function checkDefaultFill(config) {
   }
   return config;
 }
+function checkSmallestSizeTwentyTwoPixel() {
+  var twentyTwoPixel = 'normal normal 12px Helvetica';
+  var sixPixel = 'normal normal 6px Helvetica';
+  getDummyContext().font = twentyTwoPixel;
+  var twentyTwoPixelWidth = getDummyContext().measureText('Konva').width;
+  getDummyContext().font = sixPixel;
+  var sixPixelWidth = getDummyContext().measureText('Konva').width;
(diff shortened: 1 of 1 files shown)


lavrton:
With the same Chrome version, it works ok for me. What OS do you have? Can you share a screen?

coocooooo:
image

the picture of the demo

ubuntu16 chrome 73

AlfredMou:
ok, Not all chromes will have this problem.I tried windows and os chrome.The latest version of os has no problem with chrome. Old version of os and chrome on window have problems.
￼
image

lavrton:
Thanks for the Pull Request. But I am going to close it for now. I am sure the Chrome will fix that issue soon. If not - we will think to merge it. 

If such a fix is required for your app, I suggest to monkey patch `_getTextWidth` method with the updates from the Pull Request.

AlfredMou:
At the moment, I don't really need it. I can achieve smaller fonts by scaling 12px fonts in the app. Although it is more troublesome, it can guarantee font alignment.

PR #616, opened 2019-03-21
Fix: layer clear bounds should be optional

### LINKED ISSUE #615: Typescript definition not consistent with docs for clear method
Clear method argument `bounds` should be optional. The typescript definition currently requires it.

ajspera:
closes #615


lavrton:
Thanks.

PR #617, opened 2019-03-25
Examples fix

Dmitry221060:
(And GH also fixed mixed line endings)


lavrton:
Thanks a lot of the edits, but that file is generated from source code from `./src` folder. So we should edit that files instead.

PR #619, opened 2019-03-26
Initial Pointer events

jquense:
This is a quick attempt, based off of work we've done in our own app using react-konva. It's a simple emulation of pointer-events. It doesn't try and polyfill for browsers that don't support them but does implement the basic `pointerCapture` logic which makes dragging shapes a lot nicer, since you don't need to attach more handles on mouse/touch down.

I wasn't quite sure how to implement the pointer{enter|leave} logic, but would be happy to do it. I'm also not entirely sure how to test this, since there is no contributing guide but i thought i'd throw up what i have so far.


jquense:
@lavrton let me know if you think this is a good idea, and i'll be happy to finish it up :)

lavrton:
Thank looks interesting, but to be honest I don't really understand it (yet). I have no idea how to use `setPointerCapture()`.

I am thinking to make a rewrite of current event handling. So we will have true multi-touch support. But touch events will be good addition to that feature.

jquense:
`setPointerCapture` is (I think) a poorly documented feature of PointerEvents. What it does is redirect all pointer events to the element that has capture. Practically this means you can do this:

```js
onPointerEventDown(event) {
 event.currentTarget.setPointerCapture(event.pointerId)
} 

onPointerMove(event) {
  // this will fire for any pointer movement even if it's not "over" the target
}
```

This means you don't need to worry, when dragging that the element hasn't "caught up" to the pointer when moving. So you could completely replace the logic here: https://github.com/konvajs/konva/blob/master/src/shapes/Transformer.ts#L331-L333 that adds global handlers, and use `setPointerCapture` instead and the move events will work without needing to do anything.

this is a good demo using native events: https://codepen.io/patrickhlauke/pen/BKWoaa
comment out the `setPointerCapture` line to see the difference

lavrton:
I like the Pull Request. But we definitely need tests for it. Can you try to make some of them? You can take a look into the current mouse and touch tests.

jquense:
awesome yeah, I'll take a look and put up some tests 👍 thanks!

jquense:
ok @lavrton I think this is good for now. It's still missing the enter/leave events, but i could add those in a follow up. What do you think about releasing this as is, maybe without docs and i could test it out in our app?

lavrton:
Finally, I merged it but with some small changed. Currently, pointers support is optional and you have to enable it manually: `Konva._pointerEventsEnabled = true;`.

I just don't like the fact that all our internal event functions will have to run twice with pointer events.
When you move the mouse browser will trigger two events: `mousemove` and `pointermove`. So all bubbling and hit detection will be triggered twice. Probably it is ok and will work just fine. But to make some test and see how it works and I want to have the ability to explicitly enable it.

jquense:
Sounds good to me. I'll be happy to test it out and see. If it is a problem it's feasible to rework the event bubbling logic so the hit detections and bubble could be done once but fire both events

PR #622, opened 2019-03-28
Added support for holes in closed Lines. 

rfestag:
This is supported 2 ways: via fillRule (evenodd can be manually specified for self-intersecting shapes), or via a new holes parameter on Lines. Required a change to Arrow, as it inherits from Line but uses a modified ShapeConfig, and was complaining about missing parameters. Information on the two methodologies can be found here: https://stackoverflow.com/questions/13618844/polygon-with-a-hole-in-the-middle-with-html5s-canvas

I thought it made more sense to have Arrow use the LineConfig (since it inherits from Line) than ShapeConfig, as that would mean I would need to update ArrowConfig to have a `holes` property that likely isn't necessary. 

Please let me know if you have any questions or see any issues. I added two manual tests for visual verification, one using the 'evenodd' method on a self-instersecting shape, the other using multiple holes. They both have a red circle behind to demonstrate that nothing is drawn. If you see a better way to do the tests, let me know and I'd be happy to change it.

### DIFF
--- a/konva.js
+++ b/konva.js
@@ -5,10 +5,10 @@
 }(this, function () { 'use strict';
 
   /*
-   * Konva JavaScript Framework v3.2.3
+   * Konva JavaScript Framework v@@version
    * http://konvajs.org/
    * Licensed under the MIT
-   * Date: Thu Mar 21 2019
+   * Date: @@date
    *
    * Original work Copyright (C) 2011 - 2013 by Eric Rowell (KineticJS)
    * Modified work Copyright (C) 2014 - present by Anton Lavrenov (Konva)
@@ -76,10 +76,10 @@
               : {};
   var Konva2;
   (function (Konva2) {
-      Konva2.version = '3.2.3';
+      Konva2.version = '@@version';
   })(Konva2 || (Konva2 = {}));
   var Konva = {
-      version: '3.2.3',
+      version: '@@version',
       isBrowser: detectBrowser(),
       isUnminified: /comment/.test(function () {
           /* comment */
@@ -700,7 +700,7 @@
                   val = Math.round(val * 1000) / 1000;
               }
               else if (!util._isString(val)) {
-                  val = val.toString();
+                  val = val !== undefined ? val.toString() : undefined;
               }
               retArr.push(val);
           }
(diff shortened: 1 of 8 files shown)


rfestag:
I wanted to follow up on this. Would it be helpful if I removed konva.js (which appears to be built automatically during the tests) to make this easier to merge? I'm happy to make any changes you think are necessary.

lavrton:
Hello. Sorry for the late reply. I am thinking how to merge that functionality better, and thinking about possible workarounds to achieve the same visual result. `fillRule` may be confusing property. Creating holes is a very rare use case (from my point of view). 

But the main issue is that it is not universal. We may have holes for lines, but what about other shapes? Circles? Images? Can we have support here? Can we have holes with current API?

rfestag:
No worries, thanks for taking a look! As I understand, holes only make sense for any "fillable" shape. That would rule out things like images. Circles, I presume, already support something like this via rings. Other types of shapes that are more polygon-like could probably support some form of hole. I could see that being interesting, especially if the shapes had optimized hole options to support cutting out middle areas that take advantage of other parameters. For instance, take Star; we could add an additional parameter (say, fillWidth) that denotes how thick the filled region could be, and use that to create a ring equivalent to Star. I'd be happy to take a stab at that if a sticking point for you is the fact that other fillable shapes wouldn't support holes. That said, I'd definitely want your input on appropriate parameters to use for something like that. The obvious alternative is to just use the same "holes" parameter that Line supports, but I suspect anyone that wants a "Star with a hole" is probably going more for a "Star-like Ring" than a "Star with arbitrary cutout region".

I'm not aware of a good way to implement holes with the current API, but I'd be happy to play with any ideas you have. The issue is that I need to make sure some portion of the shape does not render anything at all, and I need a way to inject potentially multiple shapes to cut out (because I am using GeoJSON, and polygons are basically represented as an array of lines, where the outer shape is the first line in the array and up to N "hole" lines are specified after). I certainly don't disagree that this is probably a rare use case...the fact that this seems to be the first time it's been requested certainly demonstrates it. All I can say is that I'm loving using Konva, and found it to be a great way to quickly implement a custom renderer for GeoJSON that meets my needs. The one really missing thing for me was the ability to create holes. Most of what I end up drawing are either lines or polygons (even for things like Circles or Rings, because of projection issues at latitudes far from the equator, or other projections that aren't rectangular)

With respect to the fillRule bit, I only included it for completeness. I don't actually need that capability, and if you think it will be confusing I can pull it out. I only actually need the ability to specify hole shapes for my purposes.

lavrton:
I have to close the PR for now. But I still, want to thank you for your suggestion. I was thinking a lot how to merge that function and I didn't find a good way yet. I think that such a use case is very rare so I don't want to make bigger API for that. If you still need that functionality you can always extend `Line` shape or write your own implementation.

Another suggestion to make "holes" in the shapes is to use `globalCompositeOperation` and node caching. https://jsbin.com/rizocoluzu/1/edit?html,js,output

rfestag:
Understood, and actually I hadn't thought about adding my own custom Line. That would be a reasonable alternative for my application.

PR #630, opened 2019-04-06
Move getDummyContext into Shape class

tobymarsden:
I've been adding a new custom shape based on `Konva.Text` to support basic Markdown. All went well, and was surprisingly easy thanks to Konva's general amazingness, but I had to copy more functions from `Text` into my subclass than I wanted because the `getDummyContext` function was local to `src/shapes/Text.ts`. Because I couldn't access this dummy context from my new shape, I had to bring in every function from `Text` which called `getDummyContext()` so that it referenced the same context as the other functions in my subclass.

Very possible I've missed something here (so apologies in advance) but this PR moves the `getDummyContext` function into the `Konva.Shape` class, meaning it's now much easier to subclass `Konva.Text`, e.g. to provide a custom `sceneFunc`. 

I also removed the function from `Konva.Text` as it's now available through its parent, `Konva.Shape`.

### DIFF
--- a/src/Shape.ts
+++ b/src/Shape.ts
@@ -74,13 +74,6 @@ var linearGradient = 'linearGradient';
 var radialGradient = 'radialGradient';
 
 var dummyContext;
-function getDummyContext() {
-  if (dummyContext) {
-    return dummyContext;
-  }
-  dummyContext = Util.createCanvasElement().getContext('2d');
-  return dummyContext;
-}
 
 export const shapes = {};
 
@@ -201,6 +194,14 @@ export class Shape<Config extends ShapeConfig = ShapeConfig> extends Node<
     );
   }
 
+  _getDummyContext() {
+    if (dummyContext) {
+      return dummyContext;
+    }
+    dummyContext = Util.createCanvasElement().getContext('2d');
+    return dummyContext;
+  }
+
   /**
    * get canvas context tied to the layer
    * @method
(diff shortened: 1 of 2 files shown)


lavrton:
Thanks for the Pull Request. But to be honest I don't think it is very useful. Very rare shapes need to have access to such a context. I think it is ok to just duplicate `getDummyContext` function and use another context in your own situation.

tobymarsden:
Sure! The issue isn't needing to duplicate `getDummyContext` though, it's that subclassing `Text` is tricky as things are now because you also have to duplicate every single function which references `getDummyContext` in your subclass, because it's scoped to the file rather than the class. 

With this change, you can simply override, say, `sceneFunc`, rather than also having to copy+paste `_getTextWidth`, `measureSize` and `_setTextData`, which makes things significantly more maintainable.

lavrton:
Can show want shape you are trying to create?

Why you can not reuse `_getTextWidth`, `measureSize` and `_setTextData`? And how moving `getDummyContext` helps you?

lavrton:
Closing the issues for now as I don't see useful use cases for it.

lavrton:
Anyway, thanks for the Pull Request. Happy to discuss it again if you really need that change.

tobymarsden:
@lavrton When you try to reuse `_getTextWidth`, `measureSize` and `_setTextData` in a subclass of Konva.Shape, `getDummyContext()` ends up providing a different returned instance -- IMO `getDummyContext` should be scoped to the Konva.Shape class so you don't end up having to copy+paste every function that references it when subclassing, which is a real mess. With this change you can more easily subclass `Konva.Text` -- it makes them much more elegant.

Anyway, I have Markdown support working now -- I just have a huge chunk of duplicated code in the class because of the way `getDummyContext` is scoped, which makes maintaining subclasses in the future a pain.

lavrton:
>  When you try to reuse _getTextWidth, measureSize and _setTextData in a subclass of Konva.Shape

Reuse how? Do you have a demo or code sample?

PR #657, opened 2019-05-30
fix(pointer-events): fix possible undefined stage error

jquense:
Sometimes a shape may not have a stage associated with it? I was  hitting this in our app. I also cleaned up the event types a bit and added the pointer methods to Stage to be consistent with the other types

### DIFF
--- a/src/Node.ts
+++ b/src/Node.ts
@@ -163,21 +163,19 @@ const emptyChildren: Collection<any> = new Collection();
 let idCounter = 1;
 
 // create all the events here
-interface NodeEventMap {
+type NodeEventMap = GlobalEventHandlersEventMap & {
   [index: string]: any;
-  click: MouseEvent;
-  touchstart: TouchEvent;
-}
+};
 
-export interface KonvaEventObject<EventType extends keyof NodeEventMap> {
+export interface KonvaEventObject<EventType> {
   target: Shape | Stage;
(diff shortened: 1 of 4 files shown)


src/PointerEvents.ts
```diff
@@ -5,12 +5,10 @@ import { Stage } from './Stage';
 
 const Captures = new Map<number, Shape | Stage>();
 
-export interface KonvaPointerEvent extends KonvaEventObject<'touchstart'> {
+export interface KonvaPointerEvent extends KonvaEventObject<MouseEvent> {
```
jquense: pointer-events inherit from MouseEvent not TouchEvent

lavrton: I guess should be `PointerEvent`?

jquense: HA, yes that does make more sense

src/Node.ts
```diff
@@ -163,21 +163,19 @@ const emptyChildren: Collection<any> = new Collection();
 let idCounter = 1;
 
 // create all the events here
-interface NodeEventMap {
+type NodeEventMap = GlobalEventHandlersEventMap & {
```
lavrton: Wow. Thanks for that `GlobalEventHandlersEventMap`. Did know it exists. Super helpful.

PR #665, opened 2019-06-04
fix node registe event function type issue

iceafish:
KonvaEventListener will get a NodeEventMap entity, not NodeEventMap Key


lavrton:
Thanks!

Issue #510, opened 2018-12-04
Can not clone stage


lavrton:
Currently, it is not possible to clone a stage and reuse it. Cloning doesn't handle container change correctly.

```javascript
var cloned_stage = stage.clone();
cloned_stage.setContainer("container2");
cloned_stage.draw();
```

It this code expected result is new canvases in `container2`. But they are not here.

https://jsbin.com/kugabahoko/1/edit?html,js,output

As workaround we can convert stage into JSON and then recreate a stage from JSON.

wuhaochen:
Hi there, I got a similar issue recently when trying to use Vue-Konva to create two stages with the same vue component.

Traceback I got:
```
TypeError: Cannot read property 'removeChild' of null
    at Stage.setContainer (Stage.js?eba4:99)
    at Stage.constructor.<computed> [as container] (Factory.js?f0b4:77)
    at VueComponent.mounted (vue-konva.js?7591:308)
    at invokeWithErrorHandling (vue.runtime.esm.js?2b0e:1854)
    at callHook (vue.runtime.esm.js?2b0e:4213)
    at Object.insert (vue.runtime.esm.js?2b0e:3139)
    at invokeInsertHook (vue.runtime.esm.js?2b0e:6340)
    at Vue.patch [as __patch__] (vue.runtime.esm.js?2b0e:6559)
    at Vue._update (vue.runtime.esm.js?2b0e:3939)
    at Vue.updateComponent (vue.runtime.esm.js?2b0e:4060)
```

I think the issue is because the `parentElement` for the stage is a const `fakeContainer` and got destroyed at some point and the change here is going to try to call the `removeChild` on a `null` element.

Do you think it is reasonable to change `if (this.content)` to `if (this.content && this.content.parentElement)`?

I can try to create a minimum example if you think that is necessary.
(Basically made a Vue component including a v-stage and used it twice side by side.)

lavrton:
@wuhaochen please create a demo? Why do you need to clone the stage in `vue` app?

wuhaochen:
Hi @lavrton , I'm not able to reproduce the problem with jsfiddle or other service, however the issue shows up almost consistently when I'm trying to use `npm run serve` to test locally. Based on the Traceback, I suspect this is related to how npm update the page when source code changes.

The code I was using is basically https://codesandbox.io/s/vuekonvatemplate-d863r?fontsize=14 .

[...]

PR #668, opened 2019-06-09
Fixed set container for dangling stage, related to #510

wuhaochen:
Per our discussion in #510 .

The test I added is probably what happened to me, but I'm not sure how did it happen. The old code will such test.

The compiled file `konva.js` and `konva.min.js` has too much diff because of dependency versions etc. so I didn't include them in this patch.

### DIFF
--- a/src/Stage.ts
+++ b/src/Stage.ts
@@ -190,7 +190,9 @@ export class Stage extends Container<BaseLayer> {
     }
     this._setAttr(CONTAINER, container);
     if (this.content) {
-      this.content.parentElement.removeChild(this.content);
+      if (this.content.parentElement) {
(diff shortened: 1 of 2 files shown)


lavrton:
Thanks for the Pull Request, but probably it is not required now. There was an issue with `vue-konva`  and made a patch. Can you try the last `vue-konva` in your app? Doesn't it fix the issue you mentioned in #510?

wuhaochen:
Yes. That does fix my problem and I think the reason is exactly as stated in https://github.com/konvajs/vue-konva/issues/59. Thank you for fixing it.

On the other hand. I would still think that this change is good to make as there might be other case that similar thing could happen that resulting in a dangling stage, but I agree that it can probably wait until your next release.

PR #672, opened 2019-06-14
feat(Transformer): add draggable whitespace to anchors

DanielSchuech:
Hi,
at first thanks for the great library.

It is not easy to hit small transformer anchors on mobile devices. Big Anchors are not a real option because they are locking bad. Therefore add some free space around the visible anchors where dragging is enabled too. When using a value like 10 or 15 the feeling should be similar to most iOS apps like pages. It defaults to 0 to not change the current behaviour by default.

I currently named the property `anchorBoundingBoxSize`. If you have a better name for that property i could adjust the PR.

Best regards,
Daniel


DanielSchuech:
Fixed the tests and it is working in combination with padding now. I added the property `anchorBoundingBoxStrokeWidth` to show the draggable rect.

The draggable area for the anchors is moved 1/4 to the inside of the object and 3/4 to the outside. The following image should visualise that:
image
(blue Bbox border is only for demonstration with the parameter `anchorBoundingBoxStrokeWidth` set to 1. Default is 0)

lavrton:
Yeah, that functionality is useful, but the commit looks very complex from my point of view.

As the current workaround you can do just this:

```
transformer.find('._anchor').hitStrokeWidth(20);
```

lavrton:
I increased the default hit area for touch devices. Probably it will work just fine now.

PR #676, opened 2019-06-20
global document mock

tomasziwaszko:
There was an error while stage deserialization - document was undefined


lavrton:
Hmmm. It is better not to touch `global` variable. Can you show a demo a code sample where you see the error?

tomasziwaszko:
`Konva.Node.create(jsonData, 'container');` There is an other way to deserialize stage on Node.js side?

lavrton:
Just create a node without a container `Konva.Node.create(jsonData);`. Works good for me.

tomasziwaszko:
Works good for me too. Thx

PR #704, opened 2019-08-07
Do not stop dragging

myzonjkee:
Hi, 
I have prepared an [example](https://codesandbox.io/s/konva-drag-example-lfq3r) where you can reproduce my issue. Try to drag down a first rectangle and check your console.

In your console you may see something like this:

DevTools_-_codesandbox_io_s_tender-lehmann-lfq3r

Important part of a call stack I have emphasised by yellow color.
After call "insertBefore" it calls "Node.remove" where dragging will be braked...

### DIFF
--- a/src/Node.ts
+++ b/src/Node.ts
@@ -825,9 +825,6 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
    * node.remove();
    */
   remove() {
-    // we can have drag element but that is not dragged yet
-    // so just clear it
-    DD._dragElements.delete(this._id);
     this._remove();
     return this;
   }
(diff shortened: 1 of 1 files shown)


lavrton:
I can't reproduce. I don't see any errors. And we should stop dragging if a node is going to be removed.

myzonjkee:
Yep. 
Thank you for your reply)

lavrton:
But I am still interested to know how did you have that error. If you can reproduce it, please add more info.

myzonjkee:
The [example](https://codesandbox.io/s/konva-drag-example-lfq3r) I have created to reproduce the issue just works fine for me...

ezgif com-resize

lavrton:
That was a bug in `react-konva` fixed with the last version.
Check it: https://codesandbox.io/s/konva-drag-example-01ulf

lavrton:
Thanks for the report! It was very helpful.

myzonjkee:
Thank you!

Issue #713, opened 2019-08-21
getPointerPosition should return null when there is no _pointerPositions available : Cannot read property 'x' of undefined


cyrilmesvayn:
getPointerPosition will throw a TypeError Cannot read property 'x' of undefined when there is no pointerPosition available, because it tries to access a property of a falsy variable (pos).

Should look like this : 

```
Stage.prototype.getPointerPosition = function () {
    var pos = this._pointerPositions[0];
    if (!pos) {
        Util_1.Util.warn(NO_POINTERS_MESSAGE);
        return null;
    }
    return {
        x: pos.x,
        y: pos.y
    };
};
```

lavrton:
Merged the fix.

PR #714, opened 2019-08-22
A fix for #713

Reggino:
I've updated the signature of ```Stage.getIntersection()``` as well, to minimize the impact of the change.


lavrton:
Thanks!

PR #754, opened 2019-10-03
IFrame ts interface for animation frame

rndD:
Added description for Frame object in Animation. 
I have missed it in my work with konva.

Tell me if you are interested in getting rid of `any` in Konva.js. I have more declarations to submit.


lavrton:
Thanks!

lavrton:
@rndD Yeah, I want to remove all `any` and even use strict type in TS.

rndD:
thank you for nice instrument!

PR #756, opened 2019-10-06
Fix any in TS: part 1

rndD:
Hi!
I wanna help konva become free of `any` type. So it's the first part of a series of PRs.

### DIFF
--- a/src/Tween.ts
+++ b/src/Tween.ts
@@ -2,6 +2,7 @@ import { Util } from './Util';
 import { Animation } from './Animation';
 import { Node } from './Node';
 import { Konva } from './Global';
+import { Line } from './shapes/Line';
 
 var blacklist = {
     node: 1,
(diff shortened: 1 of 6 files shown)


src/Tween.ts
```diff
@@ -280,11 +281,11 @@ export class Tween {
         if (end.length > start.length) {
           // so in this case we will increase number of starting points
           trueStart = start;
-          start = Util._prepareArrayForTween(start, end, node.closed());
+          start = Util._prepareArrayForTween(start, end, (node as Line).closed());
```
rndD: I'm not sure here. Is it a really logic about `line`? Mb I can do a type guard helper above.

lavrton: Probably we can just wite `node.getAttr('closed')` here.

rndD: Are there any test which covers it? 
Can you tell me how can I test it?

lavrton: Related tests are here: https://github.com/konvajs/konva/blob/master/test/unit/Tween-test.js#L255

rndD:
Do I need to fix smth else?

lavrton:
Those are good updates. Thanks for your PR.

PR #773, opened 2019-10-23
Fixed zero bounds children resulting in NaN group bounds.

lwallent:
BugFix: Invalid Client Rect returned

When containers have shapes where width or height are zero they are skipped during bounds calculation. However when a container have only such children the returned bounds will be invalid as minX, minY, maxX and maxY are still undefined.

See Unit test for simple example.

### DIFF
--- a/src/Container.ts
+++ b/src/Container.ts
@@ -507,8 +507,7 @@ export abstract class Container<ChildType extends Node> extends Node<
         break;
       }
     }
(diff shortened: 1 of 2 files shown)


lavrton:
Good case. Thank you!

lavrton:
And a special thanks for the unit test! 👍

PR #825, opened 2020-01-16
Feature/intersects

ButsAndCats:
Hey @lavrton ,

It's far from complete, but just wanted to check in and get your thoughts on this approach.

Also I realised I commited the changes to konva.js I guess I wasn't supposed to do that?

this was my sandbox
```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>Test env</title>
    <style>
      html {
        height: 100%;
      }
      body {
        margin: 0;
        padding: 0;
        overflow: hidden;
        background-color: #f0f0f0;
        height: 100%;
      }
    </style>
    <script src="../../konva.js"></script>

  </head>

  <body>
    <div id="canvas"></div>
    <script>
      var width = window.innerWidth;
      var height = window.innerHeight;
      var stage = new Konva.Stage({
        container: 'canvas',
        width: width,
        height: height
      });

      stage.add(layer);

      function createShape() {
        var shape = new Konva.Rect({
          x: Math.random() * width,
          y: Math.random() * height,
          width: 30 + Math.random() * 30,
          height: 30 + Math.random() * 30,
          fill: 'grey',
          name: 'fillShape',
          draggable: true,
        });
        return shape;
      }

      for (var i = 0; i < 10; i++) {
        layer.add(createShape());
      }

      layer.add(new Konva.Line({
        stroke: 'red',
        strokeWidth: 2,
        name: 'collision'
      }))

      layer.draw();

      layer.on('dragmove', function(e) {
        var target = e.target;
        var targetRect = e.target.getClientRect();
        layer.find('.fillShape').each(function(shape) {
          // do not check intersection with itself
          if (shape === target) {
            return;
          }

          // console.log(shape.getClientRect())
          var intersection = Konva.Intersects.GetRectToRect(shape, target);
          if (intersection) {
            layer.findOne('.collision').points(intersection)
            shape.fill('yellow')
            target.fill('blue')
          }
          // do not need to call layer.draw() here
          // because it will be called by dragmove action
        });
      });
    </script>
  </body>
</html>

```

### DIFF
--- a/konva.js
+++ b/konva.js
@@ -5,10 +5,10 @@
 }(this, (function () { 'use strict';
 
   /*
-   * Konva JavaScript Framework v4.1.2
+   * Konva JavaScript Framework v@@version
    * http://konvajs.org/
    * Licensed under the MIT
-   * Date: Wed Jan 08 2020
+   * Date: @@date
    *
    * Original work Copyright (C) 2011 - 2013 by Eric Rowell (KineticJS)
    * Modified work Copyright (C) 2014 - present by Anton Lavrenov (Konva)
@@ -76,7 +76,7 @@
               : {};
   var Konva = {
       _global: glob,
-      version: '4.1.2',
+      version: '@@version',
       isBrowser: detectBrowser(),
       isUnminified: /param/.test(function (param) { }.toString()),
       dblClickWindow: 400,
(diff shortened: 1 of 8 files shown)


lavrton:
I am still not sure how to add such functionality into the core. Two reasons:

1. Users need interactions in rare use cases. But I want to keep the code size as small as possible (it is already too big).
2. `Konva.Intersects.GetRectToRect(shape, target);` looks strange to me.
    What if I don't have a Rect but a different shape?
    What about custom shapes?

ButsAndCats:
1. Yeah okay makes sense. Perhaps I can just write it as a plugin then it can be imported by people that really need it without bloating the core.

2. There will be a function for each so GetRectToCircle. I haven't thought about custom shapes yet. I guess it would be nice if there was a single function that could have any two shapes passed in and it magically figures it out. Maybe there is an entirely different approach I am missing.

lavrton:
I am going to close the PR for now.

For simple cases, we can use the combination of `getClientRect()` and simple rectangles intersection checks: https://konvajs.org/docs/sandbox/Collision_Detection.html

For complex cases and accurate intersections, we need different algorithms. And there are many of them.
As it will make the Konva code too big, I don't want to include it.

But it will be really good if we can have a solution here. As I can see it, we can have a separate package with such functionality (`npm install konva-intersection` ?). From my point of view, it should be something that can check intersections for any build-in shapes (rects, circles, stars, lines, etc) and any custom shape. Or it least makes it simple to customize the algorithms to allow custom shapes.

PR #840, opened 2020-02-03
Added Context to index-types.d.ts

akohring:
Context type is not available for use in sceneFunc on ShapeConfig


lavrton:
That can be useful. Thanks.

Issue #661, opened 2019-06-01
Rotation snap tolerance


MrNghia123:
Hi,
Thanks for great library.
The current rotation snap of the Transformer seem to have tolerance of +/- 5 degrees which is too large for application with higher accuracy. Should the snap tolerance be configurable, maybe with a rotationSnapTolerance parameter?

lavrton:
Yes. Do you want to make a pull request?

PR #861, opened 2020-02-28
fix: #661 Rotation snap tolerance

sutr90:


lavrton:
Good one. Thanks. I will make a release with it later.

Issue #906, opened 2020-05-10
Node.isVisible() returns true when node isn't displayed


nathan-muir:
Example:

 https://jsbin.com/hetuvepifu/1/edit?html,js,output

 - Hides a group
 - Adds a circle with `visible: true` to the "hidden" group
 - `circle.isVisible()`  returns true
 - circle is not displayed

Breakdown:

From what I can understand, `node.visible()` has 3 values: [1]

 -  `false` =  do not display
 - `'inherit'` = only display if parent is displayed
 - `true` = always display, even if parent is hidden.

`Node.isVisible()` should look at the 3 values for `visible` and return `true` if the object will be visible, and `false` if it will not (based on parents/inherit).

However, the `Konva.Group`'s `drawScene` will not render any children if the group is hidden [2].
eg, there could be some children not using inherit to force display, eg. `.visible(true)`.

This results in the call to `node.isVisible()` returning `true` when the node is not visible during render.

---

Possible solutions

1. Keep the current rendering behaviour, and fix `isVisible()` to treat `true` like `'inherit'`
2. Keep the current `isVisible()` behaviour and fix `Group#drawScene` to render visible children
3. Some mix, where we add a new `force` / `!important` sort of option. 

[1] https://github.com/konvajs/konva/blob/49218376e68ce8ad333ad479b3179da29bb9ddfd/src/Node.ts#L1018-L1057

[2] https://github.com/konvajs/konva/blob/49218376e68ce8ad333ad479b3179da29bb9ddfd/src/Container.ts#L331-L352

lavrton:
Yeah. I think the current behavior is confusing. And I think the first solution will be the best one.

Need to think about possible breaking changes with such an update.

PR #908, opened 2020-05-12
Make Node#isVisible() consistent with drawing behaviour

nathan-muir:
Remove the "inherit" value from Node#visible(), and treat `true` similar to "inherit".

Fixes konvajs/konva#906


lavrton:
Thanks. I am going to do the same for `isListening`.

PR #941, opened 2020-06-30
added animation option interface to replace type any. tested various …

risingBirdSong:
…options out with the .to method and the options i included in the interface were the ones that were clearly animating, excluded options that seemed to have no animation effect. included offset y and x because they do animate, though they may not be appropriate in the interface. not exhaustive, may be missing options which should be included.

### DIFF
--- a/src/Node.ts
+++ b/src/Node.ts
@@ -418,10 +418,10 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
     y -= offset;
 
     var cachedSceneCanvas = new SceneCanvas({
-        pixelRatio: pixelRatio,
-        width: width,
-        height: height,
-      }),
+      pixelRatio: pixelRatio,
+      width: width,
+      height: height,
+    }),
       cachedFilterCanvas = new SceneCanvas({
         pixelRatio: pixelRatio,
         width: 0,
@@ -645,8 +645,8 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
             if (typeof filter !== 'function') {
               Util.error(
                 'Filter should be type of function, but got ' +
-                  typeof filter +
-                  ' instead. Please check correct filters'
+                typeof filter +
+                ' instead. Please check correct filters'
               );
               continue;
             }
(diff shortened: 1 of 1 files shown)


src/Node.ts
```diff
@@ -2592,6 +2592,22 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
   }
 }
 
+interface animOptions {
```
lavrton, line 2595: I don't think we should define every property. We can just extend `ShapeConfig`, `ContainerConfig`, add some tween specific properties like "duration" and ability do define any custom property.

risingBirdSong: Thanks for taking a look. My motivation for the type was to get some intellisense and type checking inside the to method to help understand what valid options are.

I'll take a look at your suggestion thanks, it'll be good practice for me.

risingBirdSong: Hi Anton, 

this is just an idea based on your suggestion and incomplete, but is this more appropriate?

for pick type, i only picked types i thought were relevant to animation. 

This is just idea, i am only hoping for intellisense for the user.

//pick from nodeConfig those relating to animation
type animToPick = Pick<NodeConfig, "x" | "y" | "width" | "height" | "opacity" | "scaleX" | "scaleY" | "rotation" | "rotationDeg" | "offsetX" | "offsetY">
interface nodeAndFinish {
  // the values that failed travis tests
  node: Node;
  onfinish: Function;
}

interface animToSpecific {
  //and other possible methods / properties go here 
  duration?: number;
}

type animTo = animToPick & nodeAndFinish & animToSpecific;

Thank you

lavrton: Yes, something like that. But I don't think you need `Pick`, you can animate almost any property.

lavrton:
@risingBirdSong Are you interested to finish the PR?

risingBirdSong:
@lavrton Thanks but I don't think I can offer a high quality PR at this time.

risingBirdSong:
@lavrton Oh I misunderstood you, I thought you meant to write a new PR implementing the better idea that we previously discussed. Thank you

lavrton:
Yes, that what I meant. But I still merged with my own changes (from the discussion). Just to keep your contribution in history. 👍

risingBirdSong:
Oh that makes sense, nice. Thank you 👍

Issue #486, opened 2018-11-03
strokeScaleEnabled is not calculating Shape width/height correctly


bensteers:
I would expect the stroke of the rect to stay lined up with the border of the transformer when scaling, but it is shifting as if strokeScaleEnabled was not enabled.

Here's an example: https://jsbin.com/xopiyusisu/edit?html,js,output

bensteers:
after Googling around, it seems like this might just be a quirk of canvas scale transforms and the way strokeScaleEnabled is [implemented](https://github.com/konvajs/konva/blob/1c863ae389d880409f2b99e95c63620f9e0c1bf3/src/Context.js#L624), (correct me if I'm wrong). So I'm not sure if this issue easily fixable using scaling. 

One solution that I had would be to add a flag to the Transformer which, instead of using scaleX and scaleY, could directly apply the scaling to the width and height of the shape. This wouldn't fix this issue for shapes scaled in other ways, but maybe it's less of an issue for people who aren't using transformers anyway?

e.g. looking at [this line](https://github.com/konvajs/konva/blob/d15c530520c012b3304ce2625ec73ab1e69f9759/src/shapes/Transformer.js#L648)
```javascript
if(this.directScalingEnabled()) { // or whatever it should be called
  // no scaling happening here
  this.getNode().setAttrs({
    width: newAttrs.width,
    height: newAttrs.height,
    // scaleX: 1, scaleY: 1,
    x: newAttrs.x - (dx * Math.cos(rotation) + dy * Math.sin(-rotation)),
    y: newAttrs.y - (dy * Math.cos(rotation) + dx * Math.sin(rotation))
  });
else {
  // original transformation
  this.getNode().setAttrs({
    scaleX: scaleX,
    scaleY: scaleY,
    x: newAttrs.x - (dx * Math.cos(rotation) + dy * Math.sin(-rotation)),
    y: newAttrs.y - (dy * Math.cos(rotation) + dx * Math.sin(rotation))
  });
}
```
I think this would be a good feature either way, actually.

lavrton:
I don't have plans to support direct changing of width/height. That solution has too many cases and it is unclear what to do in case of `Circle` and `Group` (and some other shapes).

We will need to fix `strokeScaleEnabled` size calculation and the demo will work just fine.

glisteningly:
it has same problem with scaling Lines

maybe the transformer should attched not "out" the stroke but "on" (middle of) the stroke?  

looks like padding: -(strokeWidth/2)

[...]

PR #953, opened 2020-07-23
add `ignoreStroke` to transformer config

devrnt:
`ignoreStroke` was added in issue https://github.com/konvajs/konva/issues/486  as an option to `Konva.Transform`.

In this PR the `ignoreStroke` option is added in the `TransformerConfig` interface.


lavrton:
Thanks.

PR #962, opened 2020-08-24
Update index-types.d.ts

eduardojl:
In typescript I have an error Property "'hitOnDragEnabled' does not exist on type 'typeof Konva'." because it is not defined in this types file


lavrton:
Thanks.

Issue #958, opened 2020-08-13
[Feature Request] Add cornerRadius array feature to React Konva's Tag element


Devwulf:
Currently, the cornerRadius attribute of the Tag element for Labels currently only supports a single number to assign a radius to all corners of the tag. It would be great if a feature similar to that of Rect's cornerRadius, which also accepts an array to assign a different radius to each corner.

lavrton:
Thanks to @raphael-papazikas we have that function now. I will make a release with it later.

PR #975, opened 2020-09-06
add cornerRadius as array to label tag

raphael-papazikas:
Solves: #958


lavrton:
Thanks for other ts fixes and small changes in `Konva.Rect` validations.

PR #978, opened 2020-09-08
Default value for transformer getNodes()

alireza-bonab:


lavrton:
That makes sense. Thanks.

PR #992, opened 2020-10-09
fix miss type of dragBoundFunc

moonrailgun:
definition `this` is useful for typescript

### DIFF
--- a/src/Node.ts
+++ b/src/Node.ts
@@ -118,7 +118,7 @@ export interface NodeConfig {
   offsetY?: number;
   draggable?: boolean;
   dragDistance?: number;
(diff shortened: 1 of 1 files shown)


Aaqu:
I have problem with type
https://stackoverflow.com/questions/74137130/konva-how-to-correctly-type-dragboundfunc

moonrailgun:
> I have problem with type https://stackoverflow.com/questions/74137130/konva-how-to-correctly-type-dragboundfunc

Maybe you should import `Node` from 'konva' ?

Aaqu:
> 

```
import Konva from "konva";
import Node = Konva.Node;
```

solved the problem, thanks!

Issue #991, opened 2020-10-04
Typescript does not know that the stage / layer children are iterable


amagee:
If I have a Konva `Stage`, I would like to loop over its children with a `for..of` loop:

```
for (let layer of stage.children) {
  for (let node of layer.children) {
    // ...
  }
}
```

However for the first line I get the error: `error TS2488: Type 'Collection<Layer>' must have a '[Symbol.iterator]()' method that returns an iterator.`

and for the second line I get: `Type 'Collection<Group | Shape<ShapeConfig>>' must have a '[Symbol.iterator]()' method that returns an iterator.`

In both cases the `[Symbol.iterator]` *does* exist and the loop works just fine (presumably by virtue of the fact that Konva does `Collection.prototype = [] as any`. It just seems Typescript isn't smart enough to pick up on this.

I'm aware that I can use the `.each` function as a workaround, but ideally it would be nice not to have to do that.

lavrton:
Are you interested to make a PR?

amagee:
Yes, but it's not immediately clear to me how to fix it. Do you have any ideas? I'm happy to have a play around in any case.

lavrton:
I added the iterator type manually. Should work for many cases. https://github.com/konvajs/konva/commit/64373c7eac031938b883d424bb651be66e914e3f

If it doesn't' work, then use some workaround like `forEach` or `toArray()`

PR #993, opened 2020-10-10
fix (?) typing of Collection class

amagee:
This appears to fix https://github.com/konvajs/konva/issues/991

Everything seems to work and the tests pass (except for "radial gradient" which seems to be failing on the current master). However I'm not a Typescript expert so there may be issues I'm unaware of.

According to [this](https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work), the `Object.setPrototypeOf` solution to fix the adding of new methods to the `Array` subclass will not work in IE <= 10. I'm not sure if this project intends to support those browsers.


lavrton:
Hey, thanks for your attempt. But for now, I want to avoid using `Object.setPrototypeOf` for perf reasons (I didn't measure it).
There is a good note about `setPrototypeOf` usage here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf

I tried to add iterators types in a different way: https://github.com/konvajs/konva/commit/64373c7eac031938b883d424bb651be66e914e3f#diff-b10e04079822a0a088ccb4d434a520cfR22

It may work with some small TS config changes.

PR #994, opened 2020-10-10
fix shape getClientRect error type

moonrailgun:
Looks like the arguments of `getClientRect` is not necessary

By the way, I change parameter of `getClientRect` from `attrs` to `config`. Unify with document: [https://konvajs.org/api/Konva.Shape.html#getClientRect](https://konvajs.org/api/Konva.Shape.html#getClientRect)


lavrton:
Thanks.

PR #1002, opened 2020-10-25
export type box and others type complement

moonrailgun:


lavrton:
Thanks.

I hope one day we will enable strict mode in TS, so we can get rid of all `any`.

moonrailgun:
yep. i think we can start it right now.

PR #1012, opened 2020-11-13
Add missing text props to config type definition

tkindy:
The `fontVariant` and `textDecoration` properties were missing from the `TextConfig` interface definition despite being available in the element itself. This PR adds those two properties to the config definition so, among other things, they're available for autocomplete in the `react-konva` Typescript bindings.


lavrton:
Thanks

PR #1085, opened 2021-03-17
Fix #1084

### LINKED ISSUE #1084: toImage() filed with "Global_1.glob.Image is not a constructor"
I'm using [konva-node](https://www.npmjs.com/package/konva-node) for rendering image to PNG on the server.
```toImage()``` doesn't work for me with error "Global_1.glob.Image is not a constructor"

Quick workaround: get image URL (base64) and convert it back to the binary image with https://github.com/DiegoZoracKy/image-data-uri.

```
		const base64 = await toDataURL(stage);
		const decode = ImageDataURI.decode(base64);
		const result = decode.dataBuffer;
		const mime = decode.imageType ?? "image/png";

		res.status(200).setHeader("content-type", mime);
		res.send(result);
```

spidgorny:


lavrton:
I think it will be better to change this line https://github.com/konvajs/konva/blob/master/src/Util.ts#L669

to `var imageObj = Util.createImageElement();`

Issue #1086, opened 2021-03-18
How to detect if a Node (at least some part of it) is truly visible on the screen


AlexIII:
Hi! 
How can I check if a Node is currently visible on the screen, i.e. it is not completely outside of the boundary of the hosting canvas? It is useful for optimization, such as not drawing the parts that are falling out of user-visible area.

**Upd**
For now I solved it like this
```typescript
const isOnScreen = (testNode: Konva.Node) => {
    if(!testNode.isVisible()) return false;
    type Rect = {[k in 'x'|'y'|'width'|'height']: number};
    const haveIntersection = (r1: Rect, r2: Rect) => !(
        r2.x > r1.x + r1.width ||
        r2.x + r2.width < r1.x ||
        r2.y > r1.y + r1.height ||
        r2.y + r2.height < r1.y
    );
    const stage = testNode.getStage();
    if(!stage) return false;
    const screenRect = {
        x: -stage.x() / stage.scaleX(),
        y: -stage.y() / stage.scaleY(),
        width: stage.width() / stage.scaleX(),
        height: stage.height() / stage.scaleY()
    };
    return haveIntersection(screenRect, testNode.getClientRect({relativeTo: stage as any}));
};     
```

Is it good enough to be added in `Node` as a method? Would you be interested in PR? (Provided, there is no existing solution that I missed.)
Also notice how I had to cast `stage` to `any` in `getClientRect()` call. The typing seems to be not ok...

lavrton:
Yeah, probably it can be a part of the core. But the function has to be a bit more complicated. The most common use case is to use some "padding" in "out of the screen" calculations. So if the object is not visible on the screen, but too close to the edge it still be should be marked as visible. 

Such trick allows much more efficient checks because we need to make them less frequently.

AlexIII:
> The most common use case is to use some "padding" in "out of the screen" calculations.

Ok, that's no problem. The API will be 
```typescript
class Node {
...
isOnScreen(margin: number | {x: number, y: number}) { ... }
...
}
```
What's your minimal checklist for a PR?

lavrton:
Merged.

AlexIII:
Cool, thanks!

PR #1087, opened 2021-03-22
Add isOnScreen() method to Node #1086

AlexIII:
For #1086 . Had to also bump `rollup-plugin-typescript2` version to fix build failure.

### DIFF
--- a/src/Node.ts
+++ b/src/Node.ts
@@ -2527,6 +2527,41 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
     this.off('touchstart.konva');
   }
 
+  /**
+   * determine if node (at least partially) is currently in user-visible area
+   * @method
+   * @param {(Number | Object)} margin optional margin in pixels
+   * @param {Number} margin.x
+   * @param {Number} margin.y
+   * @returns {Boolean}
+   * @name Konva.Node#isClientRectOnScreen
+   */
+  isClientRectOnScreen(margin?: number | {x: number; y: number;}): boolean {
+    const _margin = 
+      margin === undefined
+        ? {x: 0, y: 0} 
+        : isNaN(margin as any)
+          ? margin as {x: number; y: number;} 
+          : {x: margin as number, y: margin as number}
+    ;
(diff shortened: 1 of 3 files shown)


AlexIII:
Fixed `margin`: it's now scaled too by the `Stage` scale.

src/Node.ts
```diff
@@ -2527,6 +2527,42 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
...
+  /**
+   * determine if node (at least partially) is currently in user-visible area
+   * @method
+   * @param {(Number | Object)} margin optional margin in pixels
+   * @param {Number} margin.x
+   * @param {Number} margin.y
+   * @returns {Boolean}
+   * @name Konva.Node#isOnScreen
+   */
+  isOnScreen(margin?: number | {x: number; y: number;}): boolean {
```
VahanG, line 2539: will this work for rotated rects ? May be it will be better to rename function to `isClientRectOnScreen` ? Or take in account rotation too

AlexIII: I have no idea really... It fully relies on `getClientRect()` - if it accounts for rotation then it will work, if not then it won't.
The method's name you've suggested is somewhat cumbersome for my liking, but I agree that it describes things a little more precisely. If it's between renaming the method and accounting for rotation, then I definitely for the first option. Not just because I don't want to do it, I genuinely think that's enough for the intended purpose.
Please let me know if you're ok with that and I will rename the method.

lavrton: @VahanG `getClientRect()` always returns non-rotated rectagle. But it takes rotation into account inside itself.

To me `isClientRectOnScreen` is much better. It is easy to confuse `isOnScreen()` with `isVisible()`. Also, I think we should remove `isVisible()` check from the method. So we can use the method to make the node visible.

AlexIII: Ok, it's done.
`isOnScreen()` -> `isClientRectOnScreen()`
<s>`if(!this.isVisible()) return false;`</s>

PR #1102, opened 2021-04-28
Add method to remove content event listeners

raphael-papazikas:
In my use case i need to disable for example the wheel event listener due of performance issues.

According to the current implementation the event listeners cannot be removed. 
That's why I'm opening this pull request.


lavrton:
Why wheel affects your performance?

raphael-papazikas:
@lavrton Because of this line: https://github.com/konvajs/konva/blob/master/src/Stage.ts#L858.

I use the wheel event only to zoom to position. And i dont need to check the intersections with any shapes. 
So i tried to find a way to solve these kind of issues.

lavrton:
Interesting. Do you see why does it slow for you? Is it in the FF browser?

I am not sure such an API is useful as part of the core. As a workaround you can do this:

```
stage._wheel = () => {}
```

raphael-papazikas:
Because i have many many shapes. 

@lavrton Okay thank you very much.
Closing for now

lavrton:
The number of shapes is not important for that method. 3 shapes or 30 000 should work exactly the same time. Do you have issues with other events?

raphael-papazikas:
The `getImageData` function took very long (like 50ms). So i assmued that this is caused by too many shapes.

raphael-papazikas:
@lavrton Maybe i am not seeing the issues with other events, because they are not blocking. 
But zoom to position was really laggy. 
The same applies to the drag event. I modified this as well to fit my needs.

PR #1103, opened 2021-04-29
add pointerAtEnding to Arrow

ethanruiyu:
I think this parameter will come in handy in some cases


lavrton:
Yes. It can be useful.

Issue #1106, opened 2021-05-03
[Feature request] hit canvas pixel ratio


raphael-papazikas:
Can you think of a way to provide an option for defining the pixelRatio of the hit canvas while caching? 
Or is a `pixelRatio < 1` affecting the correctness of the hit detection?

The main reason is to save memory usage when caching shapes. 
A pixel ratio of `0.5` lead to `1/4` memory usage for the hit canvas.

If you could imagine such an option inside the konva core, i could submit a PR.

lavrton:
> Or is a pixelRatio < 1 affecting the correctness of the hit detection

It may, but usually not much. 
Why do you need to save memory? Do you have a lot of caches? What do you cache?

raphael-papazikas:
Yes i have a lot of caches. 
Basically curved polygons. And for performance reasons i have to cache them. 

And it does not make sense for my use case to have the hit canvas be cached with a higher resolution than my shape itself. 
I cache the shapes with pixelRatio 0.7

raphael-papazikas:
@lavrton So if you dont mind, i would submit a PR with pixelRatio configureable

lavrton:
Yes

PR #1107, opened 2021-05-03
hit canvas pixel ratio configurable

raphael-papazikas:
Requested feature: https://github.com/konvajs/konva/issues/1106


lavrton:
Nice. Special thanks for the tests!

PR #1119, opened 2021-06-01
Tighten TypeScript definitions for shape configs

### LINKED ISSUE #1118: Make TypeScript definitions stricter
I see room for the TypeScript definitions to be improved to provide better type safety, compile time errors, and auto completion. One such example would be TagConfig's [pointerDirection](https://github.com/konvajs/konva/blob/b6e8afec6e590001fed57ff9c6a249f2fa83d06a/src/shapes/Label.ts#L179) which is typed as a string but should probably be restricted to just the few strings which are relevant `"up" | "down" | "right" | "left"`. Or another area for improvement would be the [cornerRadius](https://github.com/konvajs/konva/blob/b6e8afec6e590001fed57ff9c6a249f2fa83d06a/src/shapes/Label.ts#L182) for the same element. Instead of any number or array of numbers, there's really only 5 discrete options. Either a number, or an array with 1, 2, 3, or 4 numbers.

This will definitely require a major version bump as it'll likely cause issues with the stricter types. But most of the issues that this update will identify are likely legitimate issues which should be spotted at compile time.

I see that recently you are actually loosening the TypeScript rules which makes me concerned. I'm happy to go through this library and bring it back to a strict usage of TypeScript if this is something you are on board with. Also would probably be a good idea to add ESLint to ensure that we are as strict as possible.

CreativeTechGuy:
Fixes #1118

These types were updated based on the API Docs here: https://konvajs.org/api/Konva.html

The goal is to provide better auto-completion and compile-time errors for the interface that a user would interact with.


src/shapes/Line.ts
```diff
@@ -51,7 +51,7 @@ function expandPoints(p, tension) {
 }
 
 export interface LineConfig extends ShapeConfig {
-  points?: number[];
+  points?: [number, number, number, number, number, number];
```
raphael-papazikas, line 54: Guess this one is incorrect. 
A line can have dynamic amount of points

CreativeTechGuy: Good catch. I got deceived by the example in the docs explicitly listing 3 pairs of points.

PR #1133, opened 2021-06-24
add more accurate way to measure text width and height

veksa:
Measure text more accurate by the https://developer.mozilla.org/en-US/docs/Web/API/TextMetrics


lavrton:
Interesting. Do you have any example where it gives better results?

veksa:
prepared an example: https://jsfiddle.net/mendler/u70pnygv/

chrome, firefox and safari give different results for "height" as seen in red border, and the same for the blue border.
this is important for vertical and horizontal text centering.

lavrton:
Closing as I don't understand why we need it.

PR #1140, opened 2021-06-30
Make sure to schedule draw when calling removeChildren/destroyChildren

AlbertBrand:
When autoDraw is enabled, calling `removeChildren` or `destroyChildren` doesn't schedule a draw (because the children are detached before they are removed). This fix adds a schedule of a draw to those calls.


lavrton:
Good one. Special thanks for the tests.

PR #1143, opened 2021-07-02
feat: implement multi-styled text shape

### LINKED ISSUE #1131: feature request: multi-styled text
Thank you for this great library!

I'm trying to display multi-styled text with Konva (i.e. a text with some letters in a different color / font / style than others).

It seems that it's not possible natively with Konva (or I haven't found any configuration / node for it). 

Do you know how I can achieve this? I thought of using multiple text elements but it means that I need to re-implement text wrapping, text position, alignment...

Thank you

Eywek:
fix #1131

---

The idea of this new shape is to be able to render text with multiple styles (i.e. char 0 to 10 is blue, char 10 to 20 is black and with another font size).

I've used the Text shape and updated it to be able to support this feature. I've added more types than in the text shape (i.e. in getters/setters & context types) and tried to add all methods in the class.

To apply styles to the text you need to provide `textStyles` property for each text part. I'm not sure that was the best approach, but it was the easier in my use-case and I haven't found any other easy way to do it.

I don't know if @lavrton plan to merge this, but maybe it could help some people that need this feature.
Since the new shape use the same logic as the text shape (text wrapping, underline...) maybe it could be a good move to merge them.
I haven't done any tests (excepts manual ones, and everything seems to work) and I don't really plan to do any of them soon, I don't have really the time for it.

Anyone who want to improve this PR is welcomed (I've allowed maintainers to edit this PR but I also can add users to my fork is someone want to update the branch).

### DIFF
--- a/src/_FullInternals.ts
+++ b/src/_FullInternals.ts
@@ -17,6 +17,7 @@ import { Ring } from './shapes/Ring';
 import { Sprite } from './shapes/Sprite';
 import { Star } from './shapes/Star';
 import { Text } from './shapes/Text';
+import { MultiStyledText } from './shapes/MultiStyledText';
 import { TextPath } from './shapes/TextPath';
 import { Transformer } from './shapes/Transformer';
 import { Wedge } from './shapes/Wedge';
@@ -59,6 +60,7 @@ export const Konva = Core.Util._assign(Core, {
   Star,
   Text,
   TextPath,
+  MultiStyledText,
   Transformer,
   Wedge,
   /**
--- a/src/shapes/MultiStyledText.ts
+++ b/src/shapes/MultiStyledText.ts
@@ -1,0 +1,671 @@
+import { Context } from '../Context'
+import { Factory } from '../Factory'
+import { _registerNode } from '../Global'
+import { Shape, ShapeConfig } from '../Shape'
+import { GetSet } from '../types'
+import { Util } from '../Util'
+import { getNumberOrAutoValidator, getNumberValidator, getBooleanValidator, getStringValidator } from '../Validators'
+
+let dummyContext: Context & CanvasRenderingContext2D
+function getDummyContext() {
+  if (dummyContext) {
+    return dummyContext
+  }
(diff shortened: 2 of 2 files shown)


lavrton:
@Eywek thanks a lot for sharing your work! I am going to think how to integrate it into Konva core. For now I don't want to merge with `Konva.Text`. Because it is already too complex. And supporting rich text inside it, will make it even harder. 

But it may be a good idea for another shape.

songispm:
something new? I need it.

DavidVaness:
Any thought progress you can share on this feature @lavrton?

aokaywe:
I neet this feature.

lavrton:
Question to everybody who needs this feature. Do you need "read-only" multistyle text? Like a regular `Konva.Text` shape? Or do you expect some ability to change it with the editor?

JamieBohannaWebDev:
For me, I think the read-only would suffice. I have recreated rich text (excluding multi styles for a single text shape) formatting by listening to QuillJS for formatting changes and applying them to Konva.Text shape manually each time.

DavidVaness:
I have to agree with @JamieBohannaWebDev here, we are also using quill for the editor part.
The readOnly variant would already be really helpful to us

joaodiogocosta:
Read-only 👍

OtterlyOlive:
Any update on this? 😊 
Read-only would be very very useful!

5opr4ni:
Read only, would be most useful.

Offbeatmammal:
a read-only option to render an html/css styled block would be awesome. editing could be done using something like [popline](https://kenshin54.github.io/popline/) (similar to the 'complex' text demo) but having styled text objects would be fantastic

tjrchampion:
Yep, read only would be fantastic. html2canvas option just isn't good enough. If i wanted to use this today without it being added to konva core, how would i go about it?

lavrton:
Closed by https://github.com/konvajs/konva/issues/1131#issuecomment-4158707338

PR #1190, opened 2021-10-11
added overline text decoration support

MinimumDelta:


lavrton:
It is a good feature. Can you add a test into a pull request? Like checking trace of the context?

PR #1249, opened 2021-12-14
fix: rgbaColorToRGBA function output NaN when input with percentage

derekjan1240:


lavrton:
Can you create a test?

PR #1253, opened 2021-12-17
fix arc getSelfRect bound

lalop:
The current bound computed by Arc#getSelfRect isn't accurate.
As you can see here https://codepen.io/lalop/pen/WNZjXjy
On the first Arc the innerRadius should be used and on the last one the rect represent the bound of a angle of 270 instead of 90.


lavrton:
Can you add more tests? To make sure they fail before fixes and works ok with them?

lalop:
Sure, it's done

lavrton:
Perfect, thanks.

Issue #1296, opened 2022-03-07
Incorrect transformer bounding box with stroke and scale


testerez:
Hello,

It looks like the bounding box created around the object by a transformer is not correctly calculated when the object has both stroke and scale. I would expect it to fit the object but in some cases, it's bigger than expected. I noticed the same issue when with `getClientRect`.

Here is an example (in react): https://codesandbox.io/s/delicate-wood-bebr7i?file=/src/App.js

![Screen Shot 2022-03-07 at 14 22 39](https://user-images.githubusercontent.com/815236/157043106-002afbfb-b44f-458c-8793-ec596690856f.png)

lavrton:
For now, I am not sure how to easily fix that. Is it because of rounding of coordinates and usage of large scale.
Possible workarounds:
1. define bigger shape, use smaller scale
2. or pass `ignoreStroke` to transformer

testerez:
Thanks for your answer @lavrton. The workarounds you are suggesting may reduce the issue but the bounding box would remain off in some situations for us.
I've found that in fact the rounding and `roundingOffset` are in cause. When trying to remove it, It's working fine in the example from this issue. Here are the changes I've made: https://github.com/konvajs/konva/pull/1297
It's not clear to me why you need to round in the first place.

lavrton:
Published https://codesandbox.io/s/confident-joliot-6f5wgh?file=/src/App.js

testerez:
🥳  thanks @lavrton !

PR #1297, opened 2022-03-08
getClientRect no rounding

testerez:
This is just a draft to discuss this issue: https://github.com/konvajs/konva/issues/1296


lavrton:
Need to think how to polish that. `node.getClientRect()` was implemented to enable smooth caching. As it is cache-focused, it is doing rounding (canvas size should not be float).

Issue #1311, opened 2022-03-30
clipFunc Path2D  Turn black


3400442579:
ver：8.3.5

 _layer.clipFunc(function (ctx) {
      //ctx.arc(200, 200, 150, 0, Math.PI * 2, false);   **// Use this code to display normally**
     

**//Use the following code to display only the black circle**
      const path2D = new Path2D();
      path2D.arc(200, 200, 150, 0, Math.PI * 2, false);
      ctx.fill(path2D);

  });

lavrton:
Looks like currently Konva can't support that. We need to adapt Konva API to support that. 
https://github.com/konvajs/konva/blob/master/src/Container.ts#L401 here we should be able to pass generated `Path2d` object. I am not certain how. Return from `clipFunc`? Create another attribute?

Are you interested to think about it and make a Pull Request?

PR #1315, opened 2022-04-01
clipFunc support Path2d

3400442579:
layer.clipFunc(function (ctx) {
   const path2D = new Path2D();
   path2D.arc(left, top, r, 0, Math.PI * 2, false);
   return path2D;
 });


3400442579:
[1311](https://github.com/konvajs/konva/issues/1311)

lavrton:
Can you create another Pull Request for `Tween` fixes and remove these changes from that PR?

Issue #1304, opened 2022-03-15
textPath question


3400442579:
![image](https://user-images.githubusercontent.com/34226979/158293257-9df35104-c280-4e0f-8c6e-ef915868244d.png)

![动画](https://user-images.githubusercontent.com/34226979/158356550-e0c21837-47e8-40f5-b48b-233ac98c5bb0.gif)

path:M0,0 A464,464 0 0,1 303.5440883795784,0
test Text :对象是在画布上基于每个1
 ver:8.3.4 

```

<!DOCTYPE html>
<html>

<head>
    <script src="/lib/konva/konva.min.js"></script>
    <script src="/lib/fabric.js/fabric.min.js"></script>
    <meta charset="utf-8" />
    <title>Konva Label Demo</title>
    <link rel="stylesheet" href="/css/app.css" asp-append-version="true" />
</head>

<body>
    <table>
        <tr>
            <td >
                <input type="range" style="width: 100%;" value="2" max="100" min="-100" />
            </td>
            <td> </td>
            <td>
                path: <span id="path"></span>
            </td>
        </tr>
        <tr>
            <td>
                <span> konva</span>
                <div id="container" style="background-color:bisque;"></div>
            </td>
            <td> </td>
            <td>
                <span>fabric</span>
                <div style="background-color:bisque;">
                    <canvas id="canvas" width="600" height="600">
                </div>
                </div>
            </td>
        </tr>
    </table>

    <script>
        var s = '对象是在画布上基于每个'
        var canvas = new fabric.Canvas('canvas');
        var stage = new Konva.Stage({
            container: document.getElementById('container'),
            width: 600,
            height: 600,
        });
        var layer = new Konva.Layer();
        stage.add(layer);

        var pat=getArcPath(300, 100, 0);
        var fabricText = new fabric.Text(s, {
            fontSize: 22,
            fontFamily: 'Arial',
            top: 100,
            left: 200,
            //textAlign: 'center',
            pathAlign: 'center',
            charSpacing: 50,
            path: new fabric.Path(pat, {
                strokeWidth: 2,
                fill: null,
                stroke: '#ff0000',
                visible: true
            })
        });
        canvas.add(fabricText);
       
        var txt = new Konva.TextPath({
            x: 100, y: 200,
            draggable: true,
            fill: '#333',
            fontSize: 22,
            fontFamily: 'Arial',
            text: s,
            //align: 'center',
            //textBaseline: "top",
            letterSpacing: 5,
            data: pat
[...]

Issue #1254, opened 2021-12-21
TextPath align = center   incorrect


3400442579:
TextPath align = center   incorrect

lavrton:
Demo?

3400442579:
![image](https://user-images.githubusercontent.com/34226979/147074303-5722d269-bf0b-4dc7-9722-48b841ff1662.png)

PR #1318, opened 2022-04-06
textPath optimize (align:center and text position)

3400442579:
[bugs fixes]  align= 'center' and text position question
[1304](https://github.com/konvajs/konva/issues/1304)
[1254](https://github.com/konvajs/konva/issues/1254)

### DIFF
--- a/konva.js
+++ b/konva.js
@@ -5,10 +5,10 @@
 })(this, (function () { 'use strict';
 
   /*
-   * Konva JavaScript Framework v8.3.5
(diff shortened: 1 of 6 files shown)


3400442579:
动画

lavrton:
It is good that it works, but I can't merge this pull request. Just copying a huge code part from another library is not a good approach. 

Also, most of `PathUtil.ts` functions already presented in Konva codebase. So to me, it makes sense to only slightly modify them or add missing functions.

PR #1321, opened 2022-04-07
Fix `Cannot read properties of undefined (reading 'getWidth')`

ilyakarev:
postcss is required to run tests.
canvas in Context.ts is undefined on heavy redraws with react-konva & react-redux.
barely same issue https://github.com/konvajs/react-konva/issues/636


ilyakarev:
Did not fixed

PR #1324, opened 2022-04-11
Fix math in Transform.setAbsolutePosition

jakob0fischl:
The current implementation produces incorrect results when the current rotation of the Transform is either 90° or 270°

Incorrect Results from the current implementation:
```js
const transform1 = new Konva.Transform().setAbsolutePosition(25, 40);
// .getTranslation() => { x: 25, y: 40 }
const transform2 = new Konva.Transform().rotate(Math.PI / 4).setAbsolutePosition(25, 40);
// .getTranslation() => { x: 25, y: 40 }
const transform3 = new Konva.Transform().rotate(Math.PI / 2).setAbsolutePosition(25, 40);
// .getTranslation() => { x: 25, y: 58.020217441862236 }
const transform4 = new Konva.Transform().rotate(Math.PI / 2 * 3).setAbsolutePosition(25, 40);
// .getTranslation() => { x: 25, y: 38.68014496124148 }
const transform5 = new Konva.Transform().rotate(Math.PI).setAbsolutePosition(25, 40);
// .getTranslation() => { x: 25, y: 40 }
```

New Results:
```js
const transform1 = new Konva.Transform().setAbsolutePosition(25, 40);
// .getTranslation() => { x: 25, y: 40 }
const transform2 = new Konva.Transform().rotate(Math.PI / 4).setAbsolutePosition(25, 40);
// .getTranslation() => { x: 25, y: 40 }
const transform3 = new Konva.Transform().rotate(Math.PI / 2).setAbsolutePosition(25, 40);
// .getTranslation() => { x: 25, y: 40 }
const transform4 = new Konva.Transform().rotate(Math.PI / 2 * 3).setAbsolutePosition(25, 40);
// .getTranslation() => { x: 25, y: 40 }
const transform5 = new Konva.Transform().rotate(Math.PI).setAbsolutePosition(25, 40);
// .getTranslation() => { x: 25, y: 40 }
```

### DIFF
--- a/src/Util.ts
+++ b/src/Util.ts
@@ -207,25 +207,6 @@ export class Transform {
     return this.m;
   }
   /**
-   * set to absolute position via translation
-   * @method
-   * @name Konva.Transform#setAbsolutePosition
-   * @returns {Konva.Transform}
-   * @author ericdrowell
-   */
-  setAbsolutePosition(x: number, y: number) {
-    var m0 = this.m[0],
-      m1 = this.m[1],
-      m2 = this.m[2],
-      m3 = this.m[3],
-      m4 = this.m[4],
-      m5 = this.m[5],
-      yt = (m0 * (y - m5) - m1 * (x - m4)) / (m0 * m3 - m1 * m2),
-      xt = (x - m4 - m2 * yt) / m0;
-
-    return this.translate(xt, yt);
-  }
-  /**
    * convert transformation matrix back into node's attributes
    * @method
    * @name Konva.Transform#decompose


lavrton:
Interesting. How did you find it? Why do you use `Transform` directly?

I just found this method is not used in Konva internals. So it makes sense to remove it at all...

jakob0fischl:
`How did you find it?`
I'm currently implementing something similar to the existing Transformer for an image cropper, but functionality is different enough that I can't use that.

` Why do you use Transform directly?`
The only reason I'm using it instead of a DOMMatrix is that it has the `decompose` method which is really useful.

Updated the PR to remove the method

PR #1328, opened 2022-04-27
add possibility for top,right,bottom and left rotation anchors

biodiv:
I required to have more than one rotation anchor, because after moving a shape partly off the canvas, the rotation anchor might become invisible and therefore unusable. Adding a second (or third) rotation anchor improves this situation. This PR adds the possibility to add `rotater-top`, `rotater-bottom`, `rotater-left` and `rotater-right`. The default `rotater` is renamed `rotater-top`.

What still is missing that only `rotater-top` is shown by default if `enabledAnchors` is not passed as an option.


lavrton:
Hello, thanks a lot for your Pull Request.

I need to deeply think about merging it. I am not sure yet if such a feature is a common use case.
We may need to rethink naming of anchors because renaming in the PR will introduce breaking changes.
Also, we must have tests for that feature.

biodiv:
Thanks for sharing your thoughts. I am willing to add backwards compatibility making it pass the current tests.
I have no deeper insight into the tests of konva.js yet and cannot assess the effort to include the other rotation nodes.

lavrton:
Closing the PR for now. If we would allow moving of rotation anchor, I think we should define it with a different API. Something like "rotateAnchorPosition".

PR #1362, opened 2022-07-17
Added toBlob. Updated toImage to return a Promise.

zwc0:
Added a toBlob method which uses toCanvas internally to return a promise with the blob result. Also revised the toImage method to return a promise for the developer to optionally use promises and async/await in addition to the existing callback config method.

### DIFF
--- a/konva.js
+++ b/konva.js
@@ -3987,7 +3987,7 @@
        * @method
        * @name Konva.Node#toCanvas
        * @param {Object} config
-       * @param {Function} config.callback function executed when the composite has completed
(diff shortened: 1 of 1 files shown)


lavrton:
Hey, thanks for the PR. But I can't merge it. All changes must be done via files in `src` folder. `konva.js` file is generated from the source.

zwc0:
Thanks for the note and I apologize for the goof. I just submitted a new request off a separate branch to try to keep code/history clean. Please let me know if there are still any issues.
https://github.com/konvajs/konva/pull/1368

PR #1368, opened 2022-07-21
Added toBlob. Added Promise return to toImage. (fixed)

zwc0:
Updated the toImage method to return a promise in addition to the optional callback on the config object. Also added a toBlob method which internally uses toCanvas.


lavrton:
That is much better. Can you add tests for these functions?

zwc0:
Yes sir, I'll get those added.

zwc0:
Tests have been added to Stage-tests.ts and confirmed passed in unit-tests.html.

PR #1372, opened 2022-07-25
make sure `preventDefault()` is not called on touchend-events per default

cgrabmann:
Even though the comment states something different the current implementation of the `pointerup` event listener always calls `preventDefault()`.

This has a very negative side effect:    
On Touch-Devices the Browser still emits `click` event, but if `peventDefault()` is called on the `touchend` event the Browser does not emit a `click` event for it.    
Therefore click-listeners registered on a parent dom-element do work on Desktop, but do nothing on Mobile/Tablet.

### DIFF
--- a/src/Stage.ts
+++ b/src/Stage.ts
@@ -737,7 +737,7 @@ export class Stage extends Container<Layer> {
 
     // always call preventDefault for desktop events because some browsers
     // try to drag and drop the canvas element
-    if (evt.cancelable) {
+    if (evt.cancelable && eventType !== "touch") {
       evt.preventDefault();
     }
   }


lavrton:
Wait. That does not help. We must prevent it in touchstart.

cgrabmann:
What do you mean?

I tested it on our project and with that change the click events where emitted on the parent dom-elements on touch devices.

cgrabmann:
Hi @lavrton to elaborate on my point a bit more:

The [specification](https://w3c.github.io/touch-events/#event-touchend) for the touchend event states in the last paragraph:

> If this event is [canceled](https://w3c.github.io/touch-events/#dfn-canceled-event), any sequence of touch events that includes this event must not be [interpreted as a click](https://w3c.github.io/touch-events/#click-events).

This is exactly what I'm observing and what this change would prevent from happening by default.    
In my opinion Konva should not interfere with the default event propagation if somehow avoidable, because users of Konva would not expect that some events might not get propagated by default. Users should just call `preventDefault()` themselves if they want this to happen.

It is especially problematic in this case, because the click listener on a parent dom-element works fine on desktop, but does not work at all on touch-devices.

Regarding preventing touchstart:

I'm not 100% sure if I understood you correctly, but calling `preventDefault()` on the touchstart event would be even more detremental.

The [specification](https://w3c.github.io/touch-events/#event-touchstart) defines the following behavior for calling `preventDefault()` on the touchstart event:

> If this event is [canceled](https://w3c.github.io/touch-events/#dfn-canceled-event), it should prevent any default actions caused by any touch events associated with the same [active touch point](https://w3c.github.io/touch-events/#dfn-active-touch-point), including mouse events or scrolling.

This would prevent any conversion from touch to mouse/scroll/click events, which could lead to even more unexpected cases of events not being emitted on parent dom-elements on touch-devices.

How should I change the fix, so that you accept it?    
I would really like to get that in.

lavrton:
Sorry, you are right. Somehow when I tested your change it wasn't working. Probably I made a bad emulation. Now I tested on real device and looks like it works correctly. I added it back.

Issue #1295, opened 2022-03-07
No ellipsis added to the last line


yinguangyao:
When I set ellipsis and height, but the ellipsis is not added to the last line.
```js
const text = new Konva.Text({
  text: "1121\n22121\n432121\n",
  fontSize: 12,
  height: 30,
  width: 50,
  ellipsis: true,
  wrap: "char"
});
```
I want it to show up as
```js
1121
22121...
```
but it shows
```js
1121
22121
```

It seems that it is because the line width is smaller than the set width, which leads to the logic of else. There is no processing for ellipses here.

<img width="946" alt="image" src="https://user-images.githubusercontent.com/16210045/157000994-545a5fcf-ca00-4903-ad6a-19fff0c2d169.png">

version: 8.3.1
reproduce: https://codesandbox.io/s/unruffled-rosalind-vnytp?file=/src/index.js

yinguangyao:
Is this a bug? Can I make a pull request to fix it? @lavrton

lavrton:
It looks like a bug. It will be good to have a Pull Request with fix and test.

PR #1378, opened 2022-07-31
fix: add ellipsis for last line

yinguangyao:
When lineWidth is less than maxWidth,  add ellipses at the end of the line if necessary.

issue: [https://github.com/konvajs/konva/issues/1295](https://github.com/konvajs/konva/issues/1295)


lavrton:
Thanks for the Pull Request! I will publish new version soon.

PR #1384, opened 2022-08-08
Allow line points to be of type TypedArray.

Reggino:
For performance reasons, one may want to use a typed array when drawing lines.

This change allows this type of arrays to be used.


Reggino:
I think the failing test is not related to this change...

Issue #1396, opened 2022-09-07
The program crashes when the shadow color cannot be resolved.


iaosee:
There is an issue with setting the shadow color, the program crashes when the shadow color cannot be resolved.

When I try to enter the shadow color interactively using the input box, the program crashes if the shadow color cannot be resolved.

But it doesn't crash when fill/stroke colors can't be resolved.

[here is the live demo.  ](https://codesandbox.io/s/react-konva-bug-feedback-909dr9?file=/src/index.js)

yinguangyao:
Because this is an illegal color,  parsing RGBA fails。
But konva did not set a default RGBA color, so an error was occurs when accessing。

iaosee:
Yea, I think this error is not expected to happen, 
I expect when the color cannot be resolved, ignore it.
So I only do some processing before passing the props.

You have fixed it,  it's great! thanks.

PR #1397, opened 2022-09-10
fix: fix parsing rgba error

yinguangyao:
When parsing illegal color values, an error was occurs, because konva did not set a default RGBA color, so an error was occurs when accessing。

Issue：https://github.com/konvajs/konva/issues/1396


lavrton:
Perfect, thanks.

PR #1431, opened 2022-10-16
feat(Container): add support for empty arrays passed to `add`

MichaelDeBoey:


lavrton:
In what case you would call `add()` without children?
I am uncertain if it is a useful use case to handle.

MichaelDeBoey:
I want to use it like this:

```ts
const group = new Konva.Group(obj.attrs);

group.add(...obj.children.map(child => toKonva(child)));
```

The fact is that `obj.children` _can_ be an empty array.
If that's the case, we'll get in the case where it's calling the `getParent`, but that of course fails as `getParent` isn't available on `undefined`
https://github.com/konvajs/konva/blob/578191d5d4755b4b2d1f9490284b32d9953656ac/src/Container.ts#L127-L131

lavrton:
Closing, as I think using `container.add()` without any arguments doesn't make sense. You can check `obj.children` length before calling add.

PR #1442, opened 2022-10-31
fix: fixes canvas resources release (useful for getting rid of the canvas memory leaks on Safari macOS/iOS)

zarv1k:
Hi there!

It has been an issue for a long time in Safari, especially on iOS. This PR fixes issues with "Total canvas memory use exceeds the maximum limit (384 MB)".

Seems these are memory leaks in resources releasing for canvas elements in Safari. But Apple has been ignoring it for a while.

Releasing resources in PR made via `Utils.releaseCanvas` helper. That'll be pretty easy to remove them when Apple fixes that and it becomes unneeded.

Could you please review, merge & publish these changes.

Thank you.

### DIFF
--- a/src/Canvas.ts
+++ b/src/Canvas.ts
@@ -23,6 +23,7 @@ function getDevicePixelRatio() {
         1;
     return devicePixelRatio / backingStoreRatio;
   })();
+  Util.releaseCanvas(canvas);
   return _pixelRatio;
 }
 
--- a/src/Global.ts
+++ b/src/Global.ts
@@ -164,6 +164,17 @@ export const Konva = {
   isDragReady() {
     return !!Konva['DD'].node;
   },
+  /**
+   * Should Konva release canvas elements on destroy. Default is true.
+   * Useful to avoid memory leak issues in Safari on macOS/iOS.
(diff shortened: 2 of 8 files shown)


zarv1k:
I think there could be the new flag in `Konva` namespace, like `Konva.releaseCanvasOnDestroy`, `false` by default (turned off). On the client side, e.g. in iOS only, developer could set `Konva.releaseCanvasOnDestroy = true` to avoid these annoying issues with memory exceeding.

lavrton:
I think we need a flag, but set it true by default. I am not sure what is the case when you DON'T want to do that.

zarv1k:
Added flag

lavrton:
And one more case. Need to check `node.clearCache()` function. Cache should have three canvas elements. Need to release them too.

zarv1k:
> And one more case. Need to check `node.clearCache()` function. Cache should have three canvas elements. Need to release them too.

That's done

lavrton:
Just released. Please let me know how it works.

zarv1k:
@lavrton 
Just checked on iOS - It works like a charm. There are no more issues with memory exceeding.

PR #1456, opened 2022-11-25
Add method to toggle visibility of Node.

vikpe:
Convenience method to show/hide nodes.


vikpe:
**Samplse use case**: toggle nodes/layers when using konva as base for editing.

toggle_use_case

Pi-Bouf (approved):

lavrton:
Thanks for the PR, but I will reject it. I don't want to add non-critical functions into Konva. It is already bloated with large API.

Such function can be easily implemented by the developer.

PR #1459, opened 2022-12-04
Hex color string parsing supports four-value syntax and eight-value syntax.

iaosee:
Hex color string parsing supports four-value syntax and eight-value syntax.

https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color
Support for CSS hex color string 4/8-value syntax.


lavrton:
👍

PR #1480, opened 2023-01-12
Absolute Z-order Render Order Support

SkyeBlep:
Hello there,

Figured I should share back a feature I needed to make for my own project in case it's something you would like as well :)  I saw on Stack Overflow some people asking about it, so maybe will be useful for others.

This adds z-order support, so you can set absolute render order.  This is great for cases when you have multiple groups and want to interleave the content between them:

[example](https://github.com/SkyeBlep/konva/raw/master/AbsoluteRenderOrderGroupExample.png)

It works by adding a zOrder property to all `Node`s, and a special kind of group `AbsoluteRenderOrderGroup` that reads
and understands this property for all children.
 
Note -- in order to maintain masking behavior, cached groups are respected and treated as a single object at the group's designated z-order.  Otherwise, a group's z-order is ignored and only nodes are checked.  

## Limitations / Future Improvement

Defaults -- In future, it would probably be ideal for a node to inherit the group's z-order if not set (I haven't needed this functionality yet so I haven't written it).  For now, nodes just default to a z-order of 0.

## Code Example

```javascript
const absoluteRenderOrderGroupTest = new Konva.AbsoluteRenderOrderGroup({
	x: 0,
	y: 0
});
layer.add(absoluteRenderOrderGroupTest);

const redRect = new Konva.Rect({
	x: 0,
	y: 0,
	width: 100,
	height: 100,
	fill: 'red',
	zOrder: 10 // <------- on top
});
absoluteRenderOrderGroupTest.add(redRect);

const blueRect = new Konva.Rect({
	x: 50,
	y: 50,
	width: 100,
	height: 100,
	fill: 'blue',
	zOrder: 0 // <-------- on bottom
});
absoluteRenderOrderGroupTest.add(blueRect);
```

(There is also another example available in the tests.)

Hope it comes in handy,
-- Skye

### DIFF
--- a/src/AbsoluteRenderOrderGroup.ts
+++ b/src/AbsoluteRenderOrderGroup.ts
@@ -122,16 +122,19 @@ export class AbsoluteRenderOrderGroup extends Group {
 			});
 		} else {
 			// Is a leaf / don't descend farther -- this can be added to children map
-			let zOrder:number = node.zOrder();
+			// Determine zOrder
+			let zOrder:number = node.evaluateZOrderRecursively();
+
+			// Add zOrder if needed
 			if (!orderedChildren.has(zOrder))
 			{
 				orderedChildren.set(zOrder, new Array<Node>());
 			}
 
+			// Add to the correct bucket of zOrders
 			orderedChildren.get(zOrder).push(node); // I'd much prefer the [] syntax for clarity, but seems TS/JS doesn't seem to support it, ugh.
 		}
 	}
-
 }
 
 //AbsoluteRenderOrderGroup.prototype.nodeType = 'AbsoluteRenderOrderGroup';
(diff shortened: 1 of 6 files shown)


SkyeBlep:
> In future, it would probably be ideal for a node to inherit the group's z-order if not set (I haven't needed this functionality yet so I haven't written it)

... And then irony immediately hit me in the face and I needed that functionality, hehe.  Will update branch soon once I add it, recommend hold off for now.

SkyeBlep:
Now with zOrder inheritance added

lavrton:
Hello. Thank you so much for your pull request. Especially good that you added docs and tests. 

But I will have to decline it. I think it complicates API of the library too much. Users will be confused by a different type of group and difference between `zIndex` and `zOrder`.

Once I was thinking to add such "absolute index" feature. But I wasn't able to find a perfect design for that.

SkyeBlep:
No worries!  If anyone ends up needing the functionality, they can always just pull it from the fork

Issue #806, opened 2019-12-09
Container add more than one child cause a typescript error


bluehymn:
Container.add(node1, node2, node3...) cause following error:
**Expected 1 arguments, but got 0 or more. An argument for 'child' was not provided.**

Because the declaration file only provide one argument

```
export declare abstract class Container<ChildType extends Node> extends Node<ContainerConfig> {
    add(child: ChildType): this;

```

I checked the source code, add more than one is ok
```
 /**
   * add a child and children into container
   * @name Konva.Container#add
   * @method
   * @param {...Konva.Node} child
   * @returns {Container}
   * @example
   * layer.add(rect);
   * layer.add(shape1, shape2, shape3);
   * // remember to redraw layer if you changed something
   * layer.draw();
   */
  add(child: ChildType) {
    if (arguments.length > 1) {
      for (var i = 0; i < arguments.length; i++) {
        this.add(arguments[i]);
      }
      return this;
    }
    if (child.getParent()) {
      child.moveTo(this);
      return this;
    }
```
may need to be change to
`add(...child: ChildType[]): this;`

lavrton:
Do you want to make a Pull Request to fix the issue?

bluehymn:
> Do you want to make a Pull Request to fix the issue?

yes, I am glad to fix it

alexisfontaine:
> I checked the source code, add more than one is ok

The current signature for the `add` method is invalid. It doesn't prevent one from providing an empty array, which is not a properly handled case and raises an exception: `TypeError: Cannot read properties of undefined (reading 'getParent')`.

We could update the method signature to something like:
```ts
interface Container<ChildType> {
  add(...children: [ChildType, ...ChildType[]]): this
}
```
introducing a breaking change as the user would be required to manually check for the first element of the array:
```ts
declare const container: Container
declare const children: Node[]

const [head, ...tail] = children

if (head) {
  container.add(head, ...children)
}
```

Or, we could update the method's body to deal with empty arrays at the cost of a few CPU ticks.

lavrton:
Pull Request?

alexisfontaine:
Sure! Let me know how do you want to deal with it and I'll submit one:
1. change the type signature of the function
2. handle empty arrays at runtime

lavrton:
Personally, I would prefer the first. I think a user MUST provide at least one argument.

We can do the second, but we must check that argument is not `undefined`. I am thinking about the case:

```js
const shape = stage.find('.shape-name');
[...]

PR #1482, opened 2023-01-14
Handle adding empty children arrays to a container

alexisfontaine:
Relates to https://github.com/konvajs/konva/issues/806#issuecomment-1331891897

@lavrton, I implemented the second option. Please note that the current behavior on `undefined` arguments stays untouched (i.e. the expression `container.add(stage.find('.name'))` is still fallible).

The following expressions now execute without raising errors, in compliance with the method signature:
```ts
declare const container: Container
declare const children: Node[]

container
  .add(...children)
  .add(...[])
  .add()
```

### DIFF
--- a/src/Container.ts
+++ b/src/Container.ts
@@ -109,22 +109,27 @@ export abstract class Container<
    * add a child and children into container
    * @name Konva.Container#add
    * @method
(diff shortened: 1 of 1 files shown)


lavrton:
Thanks!

PR #1515, opened 2023-02-15
Fire pointerup on stage before click

justinhippo:


lavrton:
Can you add a test that fails before you apply your changes?

justinhippo:
> Can you add a test that fails before you apply your changes?

Added a test that checks that when we fire a click/tap it comes before a pointerup

lavrton:
Closing the PR for now, as I am not sure what really it is changing. I added tests from PR. I see `tap` after `touchend`. Only the number of taps are not equal. In PR it is 1. In the current version there are two taps. Two taps looks more correct to me. But need to check with DOM implementation.

PR #1534, opened 2023-03-18
fix: 🐛 fix types, improve type support for TypeScript

iaosee:
Improve type support for TypeScript.  
Thank you for this great library.

### DIFF
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -280,7 +280,7 @@ export class Context {
     }
   }
   _applyLineCap(shape: Shape) {
-    const lineCap = shape.attrs.lineCap;
+    const lineCap = shape.lineCap();
     if (lineCap) {
       this.setAttr('lineCap', lineCap);
     }
@@ -292,7 +292,7 @@ export class Context {
     }
   }
   _applyLineJoin(shape: Shape) {
-    var lineJoin = shape.attrs.lineJoin;
+    const lineJoin = shape.lineJoin();
     if (lineJoin) {
       this.setAttr('lineJoin', lineJoin);
     }
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -153,7 +153,7 @@ export class Context {
     }
   }
 
-  _stroke(shape) {
+  _stroke(shape: Shape) {
     // abstract
   }
 
@@ -173,7 +173,7 @@ export class Context {
     }
   }
 
-  getTrace(relaxed?, rounded?) {
+  getTrace(relaxed?: boolean, rounded?: boolean) {
     var traceArr = this.traceArr,
       len = traceArr.length,
       str = '',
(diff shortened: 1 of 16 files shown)


src/Context.ts
```diff
@@ -279,20 +279,20 @@ export class Context {
       );
     }
   }
-  _applyLineCap(shape) {
-    var lineCap = shape.getLineCap();
+  _applyLineCap(shape: Shape) {
+    const lineCap = shape.lineCap();
```
lavrton, line 283: `shape.getLineCap()` is a bit faster than `shape.lineCap()` so I would prefer keep using it instead. Not sure what to do with types here, because I don't want to define all getters and setters.

src/Context.ts
```diff
@@ -279,20 +279,20 @@ export class Context {
...
-  _applyOpacity(shape) {
+  _applyOpacity(shape: Shape) {
     var absOpacity = shape.getAbsoluteOpacity();
     if (absOpacity !== 1) {
       this.setAttr('globalAlpha', absOpacity);
     }
   }
   _applyLineJoin(shape: Shape) {
-    var lineJoin = shape.attrs.lineJoin;
+    const lineJoin = shape.lineJoin();
```
lavrton, line 295: Better to use `attrs.lineJoin` here. It is faster.

src/Context.ts
```diff
@@ -775,48 +775,48 @@ export class SceneContext extends Context {
...
+  _fillRadialGradient(shape: Shape) {
+    const grd = shape._getRadialGradient();
     if (grd) {
       this.setAttr('fillStyle', grd);
       shape._fillFunc(this);
     }
   }
-  _fill(shape) {
-    var hasColor = shape.fill(),
-      fillPriority = shape.getFillPriority();
```
lavrton, line 799: getter is a bit faster

src/Context.ts
```diff
@@ -775,48 +775,48 @@ export class SceneContext extends Context {
...
+    const hasColor = shape.fill(),
+      fillPriority = shape.fillPriority();
 
     // priority fills
     if (hasColor && fillPriority === 'color') {
       this._fillColor(shape);
       return;
     }
 
-    var hasPattern = shape.getFillPatternImage();
```
lavrton, line 807: Same getter

lavrton (changes requested):

lavrton:
@iaosee most of the changes look good. Thanks for your Pull Request.

As I mentioned in the comment, I would prefer to use getter functions instead of overloaded getter, or something even `shape.attrs` access.
These getters are not defined in types (they are generated at runtime). And I don't want to define them in types, because they should not be public. I think overloaded functions are better for end users. But inside the library itself, we should use the fastest approach.

iaosee:
@lavrton 
Thanks for the suggestion, I didn't notice this at first, I get it now.
now restored.

lavrton:
Thanks!

Issue #1006, opened 2020-10-30
Text on path not properly aligned. Wrong Chars spacing.


ramonortigosa:
Hi:
I am trying to create an export to SVG tool using konva but it seems that any text on a path is not properly aligned. The spacing between letters does not work fine too. An example:

https://jsfiddle.net/ortigosa/myj576fh/2/

<img width="607" alt="konva1" src="https://user-images.githubusercontent.com/5767717/97708559-48f1a500-1ab9-11eb-9fa7-a18afe7d1ad1.png">

What I am doing wrong?
Thanks!

lavrton:
You are doing all correctly. Our calculations are just not perfect.

ramonortigosa:
I see.. thanks for the reply.

lavrton:
Let's keep it open, so we can fix it sooner or later.

maloguertin:
> Let's keep it open, so we can fix it sooner or later.

If you want some inspiration this code seems to do a good job at spacing and centering a text on a bezier curve:

https://riptutorial.com/html5-canvas/example/19302/text-on-curve--cubic-and-quadratic-beziers

I still need to tweak it but the red text is with code inspired from this website and the black on is using html SVG text on path:


![Screenshot from 2020-11-24 15-18-41](https://user-images.githubusercontent.com/6407064/100146912-a2ef4b80-2e68-11eb-9448-9f4233bd185c.png)

lavrton:
I am going to close the issue for now, because I am not sure about the good solution for the Konva side. 
As workaround `letterSpacing` and `kerningFunc` can be used to fix it.

Issue #902, opened 2020-05-03
Align center on textPath not correctly centered


SnooHD:
See the fiddle here:
https://jsbin.com/hapipuluhi/1/edit?html,js,output

Seems like the text alignment is a bit off :(

Another thing i noticed, is that the alignment is dealing pretty weird with spaces.
https://jsbin.com/cenutasoci/1/edit?html,js,output

And afraid its also te case with letter-spacing..
https://jsbin.com/cuxojaxusu/1/edit?html,js,output

lavrton:
As I can see the demo is using `v4`. Try `v5` it works better. There were some fixes during v4.

Still not perfect for all cases, but better.

SnooHD:
Yes i see.
I know this is complicated stuff due to fonts just being a b*tch when it comes to kerning.
My intention is to have a slider where you can set a percentage of how much text will be curved.
I do wonder if you could help me thing of something where i could get this to work.. 

I have managed to get it work using SVG textPath, but i prefer using Konva as it deals with many more things i else have to reinvent the wheel for.

I will try to create a codepen with it so you have a better understanding of what i need.

SnooHD:
Ended up using codesandbox:
https://to3lm.sse.codesandbox.io/

My intention is to use this in combination with customizable fonts & text attributes like font-weight, letter-spacing etc.

gxlinhai:
I'm looking forward to it !

goldenjayr:
> Yes i see.
> I know this is complicated stuff due to fonts just being a b*tch when it comes to kerning.
> My intention is to have a slider where you can set a percentage of how much text will be curved.
> I do wonder if you could help me think of something for how i could get this to work..
> 
> I have managed to get it work using SVG textPath, but i prefer using Konva as it deals with many more things i else have to reinvent the wheel for.
> 
> I will try to create a codepen with it so you have a better understanding of what i need.
> 
> EDIT:
> Ended up using codesandbox:
> https://to3lm.sse.codesandbox.io/
> 
> My intention is to use this in combination with customizable fonts & text attributes like font-weight, letter-spacing etc.

Can you give me the source code of this example. Thanks

lavrton:
I am closing the issue as a very low priority. The issue is a hard one. But I will be happy if someone can help me with a good pull request.

PR #1539, opened 2023-03-22
Fix alignment in text path

gtktsc:
Fix glyphs alignment in text path
------------------------------------------------------------------------

Related issues:
"Text on path not properly aligned. Wrong Chars spacing. "[#1006](https://github.com/konvajs/konva/issues/1006)
"Align center on textPath not correctly centered" https://github.com/konvajs/konva/issues/902
"textpath add pathStartOffset :Offset amount for text path starting position Only used when text has a path"
 https://github.com/konvajs/konva/issues/1298
"textPath question" https://github.com/konvajs/konva/issues/1304

------------------------------------------------------------------------

This PR proposes a different way of glyphs position calculation in the `TextPath` element. 

------------------------------------------------------------------------

List of changes:
1. The new algorithm used to calculate position,
2. The new `text-paths.html` interactive example demonstrating curved text in konva,
3. Update in the `readme` file,
4. Updates in tests (more in the comments section).

------------------------------------------------------------------------

Before (current master):

This PR:

------------------------------------------------------------------------

Questions:
`Path` methods - `getLength` and `getPointAtLength` are inaccurate compared to native SVGPathElement counterparts. That is why I've decided to use native methods instead - since these might be unavailable on the server/node environment, I'm using `Konva.Path.getTotalLength` and `Konva.Path.getPointAtLength` as a fallback. Is that acceptable?


README.md
```diff
@@ -181,7 +181,7 @@ Run `npx gulp api` which will build the documentation files and place them in th
 # Pull Requests
 
 I'd be happy to review any pull requests that may better the Konva project,
-in particular if you have a bug fix, enhancement, or a new shape (see `src/shapes` for examples). Before doing so, please first make sure that all of the tests pass (`gulp lint test`).
```
gtktsc, line 184: I believe gulp lint test are no longer available? Or did I missed something?

package.json
```diff
@@ -50,7 +50,7 @@
     "test:build": "parcel build ./test/unit-tests.html --dist-dir ./test-build --target none --public-url ./ --no-source-maps",
     "test:browser": "npm run test:build && mocha-headless-chrome -f ./test-build/unit-tests.html -a disable-web-security",
     "test:node": "ts-mocha -p ./test/tsconfig.json test/unit/**/*.ts --exit && npm run test:import",
-    "test:watch": "rm -rf ./parcel-cache && parcel serve ./test/unit-tests.html ./test/manual-tests.html ./test/sandbox.html",
+    "test:watch": "rm -rf ./parcel-cache && parcel serve ./test/unit-tests.html ./test/manual-tests.html ./test/sandbox.html ./test/text-paths.html",
```
gtktsc, line 53: I've added it just for testing purposes. It's an interactive demo of the TextPath.

test/unit/Text-test.ts
```diff
@@ -667,10 +667,15 @@ describe('Text', function () {
...
 
-    assert.equal(layer.getContext().getTrace(), trace);
+    if (!isBrowser) {
+      trace =
+        'clearRect(0,0,578,200);save();transform(1,0,0,1,10,10);font=normal normal 30px Arial;textBaseline=middle;textAlign=left;translate(0,0);save();fillStyle=black;fillText(Y,0,15);fillStyle=black;fillText(O,20,15);fillStyle=black;fillText(U,43,15);fillStyle=black;fillText( ,65,15);fillStyle=black;fillText(A,73,15);fillStyle=black;fillText(R,93,15);fillStyle=black;fillText(E,115,15);fillStyle=black;fillText( ,135,15);fillStyle=black;fillText(I,143,15);fillStyle=black;fillText(N,151,15);fillStyle=black;fillText(V,173,15);fillStyle=black;fillText(I,193,15);fillStyle=black;fillText(T,201,15);fillStyle=black;fillText(E,219,15);fillStyle=black;fillText(D,239,15);fillStyle=black;fillText(!,261,15);restore();restore();';
+    }
+
+    assert.equal(layer.getContext().getTrace(false, true), trace);
```
gtktsc, line 678: I think this is unrelated to my change - there are differences in rounding in the browser/node. I've also added `getTrace(false, true)` to fix it.

test/unit/TextPath-test.ts
```diff
@@ -277,7 +277,7 @@ describe('TextPath', function () {
     layer.add(textpath);
     stage.add(layer);
 
-    cloneAndCompareLayer(layer, 200);
+    cloneAndCompareLayer(layer, 200, 10);
```
gtktsc, line 280: This one is interesting, here are screenshots from master:




Here comes this PR:






Adding a second tolerance value fixes it. I believe it is retry, right? It this acceptable?

test/unit/TextPath-test.ts
```diff
@@ -757,9 +757,9 @@ describe('TextPath', function () {
 
     // just different results in different envs
     if (isBrowser) {
-      assert.equal(Math.round(rect.height), 329, 'check height');
-    } else {
       assert.equal(Math.round(rect.height), 331, 'check height');
+    } else {
```
gtktsc, line 761: I think this is expected - the text path is rendered differently.

README.md
lavrton: gulp is here. But I don't think that `gulp lint test` command is the last one to use. it should be just `npm run test`

src/shapes/TextPath.ts
```diff
@@ -105,6 +107,21 @@ export class TextPath extends Shape<TextPathConfig> {
     this._setTextData();
   }
 
+  _readDataAttribute() {
+    this.dataArray = Path.parsePathData(this.attrs.data);
+    // in case document is not defined (server side rendering)
+    // use the KonvaJs Path class instead of the browser's native SVGPathElement
+    if (typeof window !== 'undefined' && document) {
+      this.path = document.createElementNS(
```
lavrton, line 115: That is interesting. How hard may it be to replicate native API. So we have some calculations in place, but they are not perfect, right? Are there any libs that may help?

src/shapes/TextPath.ts
```diff
@@ -105,6 +107,21 @@ export class TextPath extends Shape<TextPathConfig> {
...
+        'http://www.w3.org/2000/svg',
+        'path'
+      ) as SVGPathElement;
+      this.path.setAttribute('d', this.attrs.data);
+    } else {
+      this.path = new Path({ data: this.attrs.data }) as Path;
```
lavrton, line 121: I would think how to avoid creating an instance of the path. Probably we can somehow just use methods from it directly. It will save memory a bit. But more important, shapes must be properly destroyed. So instead of thinking about how to destroy shapes, it may be better not to create them at all.

src/shapes/TextPath.ts
lavrton: With your last fixes it still doesn't work good as native?

README.md
gtktsc: Good, so I'll keep this change 👍

src/shapes/TextPath.ts
gtktsc: This PR fixed the alignment problem, but this is a different one. 

Approximation accuracy is not crucial for this fix. Using Konva `Path.getPointAtLength` and `Path.getLength` works ok-ish, but when the path has a cubic or a quadratic Bezier curve, the letter spacing between two glyphs is inconsistent, cause the length between two points is not accurate.

I haven't touched the `Path` methods. The problem is in `Path.getPointAtLength`, probably somewhere in `getPointOnCubicBezier` or `getPointOnQuadraticBezier`. Perhaps, browsers are not using De Casteljau's algorithm, but it's just a guess.

I'm not aware of any libs having similar to the native SVG `getPointAtLength` method.

src/shapes/TextPath.ts
gtktsc: Good point. All I need in this PR is `getLength` calculation and get `getPointAtLength`.
The `getLength` can be replaced with: 
```js
this.pathLength = 0;
  for (var i = 0; i < this.dataArray.length; ++i) {
    this.pathLength += this.dataArray[i].pathLength;
}
```
Because this is how `Path` calculates it in `constructor`. 

`getPointAtLength` -  I think it is possible to create a new static method that accepts `length` and `dataArray` as a second parameter. Then use a static method in TextPath (and in `getPointAtLength` too). It should do the trick, right?

lavrton: Sounds good to me.

src/shapes/TextPath.ts
gtktsc: I found one polyfill [svg-path-properties](https://github.com/rveciana/svg-path-properties). WDYT about fixing `getPointAtLength` as a follow up PR? I prefer to keep this pull request about TextPath only.

src/shapes/TextPath.ts
gtktsc: @lavrton WDYT about this approach?

gtktsc:
Hey @lavrton, is there anything I can improve in this PR? Are there any additional steps required to merge this fix? Thanks

lavrton:
Hey, sorry for the silence. Just having busy days and a vacation. In general, PR looks excellent to me. I will try to find a time soon and make a review again and do a couple tests locally.

lavrton:
I found the case where I will probably not be able to merge PR as is.

This is "check bad calculations" test case. Old result:

Screenshot 2023-03-31 at 19 51 21


New version:

Screenshot 2023-03-31 at 20 10 25

Context of the text - it is trying to draw `__________` along the path. Visually it should look like a continuous line.
I will try to update the test to make it obvious.

gtktsc:
Hey @lavrton, thank you for the update. It appears that the root cause of the issue was related to the difference between Konva's "getPointOnCubicBezier" and the browser's "getPointOnPath" implementations. To address this, I utilized the code from the "svg-path-properties" library, which enabled me to accurately calculate the point position on the Bezier curve. As a result, the calculations between the browser and node are now consistent, with reduced discrepancies between the built-in method and the library. Furthermore, I have added a test to ensure calculation accuracy. I appreciate your time and consideration on this matter. Please let me know your thoughts.

lavrton:
I merged PR. Thanks for it.

But right now, I am thinking it will be better to rewrite it a bit. In browser we can use `SVGPathElement` for path calculations, but we need to make sure this API is available in all browsers.

For nodejs, we will make overwrite to use pure js. So we don't have to ship all aproximation code to the browser.

gtktsc:
Hey @lavrton, thank you very much!

>But right now, I am thinking it will be better to rewrite it a bit. In browser we can use SVGPathElement for path calculations, but we need to make sure this API is available in all browsers.

[As was noted earlier](https://github.com/konvajs/konva/pull/1539#issuecomment-1492772415), this could lead to differences between node and browser rendered images. I'm not convinced if this is desired behavior. In my opinion, it might be worth utilizing the [svg-path-properties](https://github.com/rveciana/svg-path-properties) library as a polyfill for calculations to keep the results consistent across all browsers and environments. However, I am uncertain if it is acceptable to use 3rd party library as a dependency for the project, but I believe it is worth exploring further. I am happy to create a pull request to demonstrate how it can be used.

Thank you again for your time and support.

PR #1552, opened 2023-04-16
extend transformer.anchorSize to support an array of width and height

mindofbp:
I have come across use cases where it would be nice to have the transformer anchors width and height be set independently of one another. 

This attempts to allow that by modifying the `Transformer#anchorSize` GetterSetter to support an array of size two, specifying the anchor width and height when used. Passing a single number still sets it to a square with width height values equal to the parameter value. I don't _think_ this introduces breaking changes but if so (and this provides value) I am happy to rethink a solution.

Thanks!

### DIFF
--- a/src/shapes/Transformer.ts
+++ b/src/shapes/Transformer.ts
@@ -6,7 +6,7 @@ import { Rect } from './Rect';
 import { Group } from '../Group';
 import { ContainerConfig } from '../Container';
 import { Konva } from '../Global';
-import { getBooleanValidator, getNumberValidator } from '../Validators';
+import { getBooleanValidator, getNumberValidator, getNumberOrArrayOfNumbersValidator } from '../Validators';
 import { _registerNode } from '../Global';
 
 import { GetSet, IRect, Vector2d } from '../types';
@@ -28,7 +28,7 @@ export interface TransformerConfig extends ContainerConfig {
   anchorFill?: string;
   anchorStroke?: string;
   anchorStrokeWidth?: number;
-  anchorSize?: number;
+  anchorSize?: number | Array<number>;
   anchorCornerRadius?: number;
   keepRatio?: boolean;
   centeredScaling?: boolean;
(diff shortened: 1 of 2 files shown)


lavrton:
What is the use case? Can you provide a visual example?

lavrton:
I see. Thank makes sense. But I am thinking to make a very different API for that. There are many requests to add some kind of customizations for anchors. Initially, I added some most obvious attrs like `anchorFill`, `anchorSize` and so on. But it is not enough, and users (like you) want to customize more. My current idea is to make universal `anchorStyleFunc` API:

```js
const tr = new Konva.Transformer({
   anchorStyleFunc: (anchor) => {
     if (anchor.hasName('middle-left')) {
       anchor.setAttrs({
         width: 10,
         height: 50
       })
     }
   }
});
```

So, a user has full control of any properties.

mindofbp:
I like this. This would give the user control over the anchor with all the methods available to `node`, yes? In this scenario, would the current styling methods be deprecated? It might be hard to handle the priority of methods if a user does something like
```
const tr = new Konva.Transformer({
   anchorStyleFunc: (anchor) => {
     if (anchor.hasName('middle-left')) {
       anchor.setAttrs({
         width: 10,
         height: 50
       })
     }
   },
   anchorSize(10),
});
```
Unless it is documented that `anchorStyleFunc` is called after the existing methods, and supplements those by not having a default?

lavrton:
Demo of what you want with the new version: https://jsbin.com/jagahifame/edit?html,js,output

```js
const tr = new Konva.Transformer({
  nodes: [shape],
  anchorStyleFunc: (anchor) => {
    if (anchor.hasName('middle-left') || anchor.hasName('middle-right')) {
      anchor.setAttrs({
         height: 20,
         offsetY: 10
      })
    }
  }
})
```

You should be just careful will offset if you change the size. Basically offset should be half of the size.

PR #1558, opened 2023-04-25
feat: :sparkles: add EllipseRing shape

iaosee:
I want to draw an ellipse ring, but I found that Konva.Ring cannot accomplish this (unless I use scaleX / scaleY, or use a custom implementation of Konva.Shape). 
Therefore, I want to add an EllipseRing shape component.

image


lavrton:
Thank you so much for the Pull Request. But I am going to close it, as I think that shape is too rare use case and I would like to keep the library minimal where possible.

It may be a good idea to have a separate collection or custom shapes or just a tutorial/doc on the website if you like to make a pull request for that.

PR #1560, opened 2023-04-28
 Issue when navigating from a Konva-enabled page to another page

andy820621:
I encountered an issue when navigating from a Konva-enabled page to another page: "Cannot read properties of undefined (reading 'setAttrs')". 

After some investigation, I found that this may be caused by the absence of an anchor (such as 'top-left', 'top-center', etc.) in my Transformer, which causes an error in the code 'anchor.setAttrs(attrs)'. 

Therefore, I would like to add an if statement to check for the existence of the anchor before setting its attributes.


lavrton:
Can you somehow make a demo of this?

PR #1574, opened 2023-06-09
fix: type improve

IcetCode:
This PR has made the following changes:
- Added missing TypeScript types
- Replaced 'var' with 'const' or 'let'
- Simplified the condition statements of the `setLayers`


lavrton:
Thanks.

PR #1600, opened 2023-07-08
✨ add color curve filter

ourfor:


ourfor:
This filter allows you to define a color-value map in each channel

ourfor:
> Isn't it the same as RGBA filter that we have?

This filter allows you to define a color-value map in each channel

quarksb:
it's doesn't make much sence to add a custom filter to konva, which would make konva heavier

PR #1609, opened 2023-07-17
Added null checks and types

razw22:


lavrton:
What was your motivation for that PR?

razw22:
> 

I use this library, and I have validators that check for types and null checks so I had to add those.
I wanted to upstream my changes so it will be available next time I'll need to update the konva version I use, and also I believe it would help to have those checks in general.

lavrton:
I am not sure if I like this PR. `!` operator in Typescript is a kind of workaround for cases, when you are telling to compiler "hey, I am sure this value is not null, don't show type error here!". I think it will be better to slightly refactor code in such cases, so the compiler is confident that the value is not null-like.

razw22:
> I am not sure if I like this PR. `!` operator in Typescript is a kind of workaround for cases, when you are telling to compiler "hey, I am sure this value is not null, don't show type error here!". I think it will be better to slightly refactor code in such cases, so the compiler is confident that the value is not null-like.

I know it's a workaround but it doesn't affect the code. The added null assertions are used where there's already a reference to a value that might be null which can cause an error.

Going forward it will be better to fix those references. However, for now it's better, so we can actually be aware where a reference might be null. It will help in ultimately making the code null safe.

lavrton:
I just updated the repo with some strict rules from typescript: https://github.com/konvajs/konva/commit/d33b8e944bd92a0f27c9bf7f6156db28c98eb77b

I think the better way to improve the typescript code base is to slowly apply more and more strict rules from typescript in `tsconfig.json`. And then just fix all the errors from the compiler. 
https://github.com/konvajs/konva/commit/d33b8e944bd92a0f27c9bf7f6156db28c98eb77b#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R12

@razw22 thank you so much for our original effort, but I will have to close it for now because there is no null errors from typescript right now. It will be good if you can improve codebase more with more rules.

PR #1624, opened 2023-08-09
Updated parameter names in context.ts to match MDN parameter names

joshrudesill:


joshrudesill:
Need to figure out tests

PR #1626, opened 2023-08-11
feat: Change Transformer transform event position

ouronghuang:
If I wang to add some node to [Transformer](https://konvajs.org/api/Konva.Transformer.html) and  binding events `transformer.on('transform')` it can not set correct position, the add node will be shake, so I adjusted the position of the event `this._fire('transform')`, it work correct.


lavrton:
What exactly does it fix? Can you make a test?

lavrton:
@ouronghuang can you provide code of your demo? I wasn't able to reproduce that.

Issue #1251, opened 2021-12-14
RTL language support for toDataURL


kanaabe:
Hi, thanks for this library. It's been very useful so far! 

One area I am having trouble with is dealing with RTL languages. I'm able to get the canvas to render RTL languages with `dir="rtl"` by directly accessing the canvas element as was suggested [here](https://konvajs.org/docs/sandbox/Native_Context_Access.html#page-title). However, I need to export that canvas as well. From the docs I see: 

> Because Konva has full control over drawing and may reset your manual drawings or lost them on any export such as stage.toDataURL().

Do you have any alternate suggestions if we need to use `stage.toDataURL`? Or is the solve to support more attributes such as `dir` as first-class attributes? I'd be happy to open a PR if that's something you'd be interested in. 

Example of it affecting RTL output: 
https://codesandbox.io/s/musing-ellis-yh005?file=/src/index.js

lavrton:
I don't have any simple solutions in mind. `node.toDataURL()` creates a temporary canvas to make an export.

As workaround you can monkey-patch `Konva.Util.createCanvasElement()` and set `dir` attribute in it.

kanaabe:
Thanks for the response @lavrton, I wanted to follow up.

Do you have any examples of this patching with `node.toDataURL()`? I tried patching and I can see the console.logs applying the `dir="rtl"` but my output canvas from `node.toDataURL()` still applies the incorrect attribute. Do you spot anything wrong with my implementation? https://codesandbox.io/s/relaxed-rumple-hkfpw?file=/src/index.js

Ultimately I want to apply a variety of attributes including CSS styles such as filters to produce the canvas the user wants, then export it. (I tried using the Konva Filters but the performance wasn't great since we are working with large images and a slider UI). My question is, if we go with patching, would we lose control over which layer we want the attributes + styles to apply to? For example, if we had two pieces of text in English and Hebrew and patched `createCanvasElement`, the style would apply to the entire canvas instead of say, the Hebrew text layer only. Is that correct? Do you have any other suggestions on how to approach this?

lavrton:
I don't know why `dir="rtl"` is not working. I'm not very knowledgeable about it, so I don't have any solutions for you. You may need to debug on your own.

> Ultimately I want to apply a variety of attributes including CSS styles such as filters to produce the canvas the user wants

[...]

PR #1637, opened 2023-09-09
feat: Add basic support for RTL text direction.

xkxx:
Currently, supporting RTL languages requires workarounds as described in https://github.com/konvajs/konva/issues/1251, which has many caveats such as not being able to support `toDataUrl` or mixing LTR and RTL languages.

This PR upstreams a basic feature addition that we have to address these cases. A new property `direction` is added on the `Text` element, which is passed through to the 2d canvas during rendering.

This PR is basic to a minimum to ensure that it doesn't disrupt LTR languages, as such I'm sure there are still edge cases for RTL that it doesn't address. I wanted to keep the initial change small so that we can iterate, while this first step should still be useful for a lot of common cases that's useful to the Konva community.

Best,

xkxx

### DIFF
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -68,7 +68,6 @@ var COMMA = ',',
     'strokeText',
     'transform',
     'translate',
-    'trySetLetterSpacing',
   ];
 
 var CONTEXT_PROPERTIES = [
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -68,6 +68,7 @@ var COMMA = ',',
     'strokeText',
     'transform',
     'translate',
+    'trySetLetterSpacing',
   ];
 
 var CONTEXT_PROPERTIES = [
@@ -92,6 +93,11 @@ var CONTEXT_PROPERTIES = [
 ] as const;
 
 const traceArrMax = 100;
+
+interface CanvasRenderingContext2DFeatureDetection extends CanvasRenderingContext2D {
+  letterSpacing: string | undefined;
+}
+
 /**
  * Konva wrapper around native 2d canvas context. It has almost the same API of 2d context with some additional functions.
  * With core Konva shapes you don't need to use this object. But you will use it if you want to create
@@ -705,6 +711,21 @@ export class Context {
    */
   translate(x: number, y: number) {
     this._context.translate(x, y);
+  }
+  /**
+    * Set letterSpacing if supported by browser.
+    * @method
+    * @name Konva.Context#trySetLetterSpacing
+    * @returns true if successful, false if not supported.
+    */
+  trySetLetterSpacing(letterSpacing: number): boolean {
+    var context = this._context as CanvasRenderingContext2DFeatureDetection;
+    var letterSpacingSupported = 'letterSpacing' in context;
+
+    if (letterSpacingSupported) {
+      context.letterSpacing = letterSpacing + "px";
+    }
+    return letterSpacingSupported;
   }
   _enableTrace() {
     var that = this,
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -78,6 +78,7 @@ var CONTEXT_PROPERTIES = [
   'shadowBlur',
   'shadowOffsetX',
   'shadowOffsetY',
+  'letterSpacing',
   'lineCap',
   'lineDashOffset',
   'lineJoin',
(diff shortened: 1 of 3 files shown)


lavrton:
That is a good PR. Can you add any unit tests and possible visual tests?

The next major and complex edge case is when text has "letterSpacing".

lavrton:
Ah, Safari... I love you so much.

For now, I suggest assuming that `letterSpacing` is not supported at all and render RTL text without it. From my point of view, consistent result across different browsers is critical.

lavrton:
The issue mentioned in pixi is already fixed in Konva.

I just afraid that polyfill will be inconsistent with the native implementation. So I prefer to use polyfill everywhere for now. I think VERY VERY deep testing will be required to make sure all possible text variations look the same in polyfill vs native. 

But, as soon as Safari will have, we should drop polyfill at all. Are there any news about such support?

We can use native `letterSpacing` for RTL, because polyfill doesn't work for it at all, right?

src/shapes/Text.ts
```diff
@@ -303,11 +315,13 @@ export class Text extends Shape<TextConfig> {
           lineTranslateX += this.measureSize(letter).width + letterSpacing;
         }
       } else {
+        context.trySetLetterSpacing(letterSpacing);
```
lavrton, line 318: I suggest to set letter spacing similar to all other attributes. Like `context.setAttr('letterSpacing', value);` or direct set.

src/Context.ts
```diff
@@ -705,6 +712,21 @@ export class Context {
   translate(x: number, y: number) {
     this._context.translate(x, y);
   }
+  /**
+    * Set letterSpacing if supported by browser.
+    * @method
+    * @name Konva.Context#trySetLetterSpacing
+    * @returns true if successful, false if not supported.
+    */
+  trySetLetterSpacing(letterSpacing: number): boolean {
```
lavrton, line 721: I think we can get rid of that method and set property dirrectly.

src/shapes/Text.ts
```diff
@@ -303,11 +315,13 @@ export class Text extends Shape<TextConfig> {
...
         this._partialTextX = lineTranslateX;
         this._partialTextY = translateY + lineTranslateY;
         this._partialText = text;
 
         context.fillStrokeShape(this);
+        context.trySetLetterSpacing(0);
```
lavrton, line 324: I don't think this like is useful. Konva should already reset any changes of the context with `save()` and `restore()`

src/shapes/Text.ts
```diff
@@ -201,12 +207,15 @@ export class Text extends Shape<TextConfig> {
     var lineTranslateX = 0;
     var lineTranslateY = 0;
 
+    context.setAttr('direction', direction);
```
lavrton, line 210: Need to think a bit about this line. Probably, we can skip it if the direction is `ltr`. Otherwise, we may need to fix many tests.

src/shapes/Text.ts
```diff
@@ -682,6 +697,22 @@ Factory.overWriteSetter(Text, 'width', getNumberOrAutoValidator());
...
+ * @param {String} direction
+ * @returns {String}
+ * @example
+ * // get direction
+ * var direction = text.direction();
+ *
+ * // set direction
+ * text.direction('rtl');
+ */
+Factory.addGetterSetter(Text, 'direction', 'ltr');
```
lavrton, line 714: Should it be `inherit` by default?

lavrton (changes requested):

src/Context.ts
```diff
@@ -68,6 +68,7 @@ var COMMA = ',',
     'strokeText',
     'transform',
     'translate',
+    'trySetLetterSpacing',
```
lavrton, line 71: You can remove this, right?

src/shapes/Text.ts
lavrton: For now, I think inherit will work just fine.
There may be a global page option for direction. So I think Konva should preserve it when possible. 

To detect what direction to apply, we can probably use this:
```js
const textDir = this.direction();
const canvasDir = canvas.direction;
const direction = textDir === 'inherit' ? canvasDir : textDir;
// then do rest of the logic in sceneFunc
```

lavrton:
Thanks for the Pull Request and your work with the feedback.

lavrton:
Released.

PR #1659, opened 2023-10-30
Fix position jitter when flip enabled false

MadeByMike:
Rather than disabling the exiting updates when flipEnabled is false I've updated to translate and set scale to an absolute value on the new transform.

This appears to solve the jitter and work otherwise the same as the existing method.

### DIFF
--- a/src/shapes/Transformer.ts
+++ b/src/shapes/Transformer.ts
@@ -1056,17 +1056,19 @@ export class Transformer extends Group {
     oldTr.scale(oldAttrs.width / baseSize, oldAttrs.height / baseSize);
 
     const newTr = new Transform();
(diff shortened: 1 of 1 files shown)


lavrton:
Tests are not passed...

MadeByMike:
The order or rotation matters :)

Test should be passing now.

lavrton:
Good. @MadeByMike is it possible to make a test that is failing, before the fix. I am a bit not sure what exactly it is fixing.

lavrton:
Actually, I just added it. I need to test it a bit more. But looks like it works much much better now. Thanks a lot @MadeByMike

MadeByMike:
No problem, happy to help 👍

PR #1667, opened 2023-11-09
add additional data parameter to node config

fveltmann:
#1666


lavrton:
I don't want to add any new attributes into types, that are not part of rendering process or main Konva logic flow.

You can always just do this:

```js
node.setAttr('additionalData', data);

node.getAttr('additionalData');
```

Issue #1672, opened 2023-11-20
[Typescript] Export GetSet types to the outside world.


StringKe:
There are currently no externally exported GetSet types. I need to use it to refine my overall type definition.

`Validators` and `Factory` and `GetSet` need to be exported.

```ts
import Konva from 'konva';

import { type ArtboardConfigWithDefault } from '../types';

export declare type ArtboardConfig = {
    artboard: ArtboardConfigWithDefault;
} & Konva.StageConfig;

export class ArtboardStage extends Konva.Stage {
    artboard: Konva.GetSet<ArtboardConfigWithDefault, this>;
    
    constructor(config: ArtboardConfig) {
        super(config);
    }
}

Konva.Factory.addGetterSetter(ArtboardStage, 'artboard', {
    wdith: 800,h eight:600
});
```

PR #1673, opened 2023-11-20
export Factory Validators and GetSet

StringKe:
Makes it easier for users to extend any element in Konva, or to create new elements.

fix: #1672

---
https://github.com/konvajs/konva/blob/master/src/index-types.d.ts

Perhaps types should be exported directly, rather than using namespace, which is currently difficult to extend.

konva.export.ts
```ts
export { Container, ContinareConfig } from "./core"
```

index.ts
```ts
import * as Konva from "./konva.export"

export default Konva
```

### DIFF
--- a/src/_CoreInternals.ts
+++ b/src/_CoreInternals.ts
@@ -2,6 +2,9 @@
 import { Konva as Global } from './Global';
 
 import { Util, Transform } from './Util';
+import { Factory } from './Factory';
+import * as Validators from './Validators';
+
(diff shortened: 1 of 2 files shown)


lavrton:
I am not sure I want to move this API into public global `Konva` object. Can you just import in directly in your project?

```js
import * as Validators from 'konva/lib/Validators';
```

StringKe:
That works. Thank you.

PR #1690, opened 2023-12-19
feat: add flag for link line visiable for Transformer's rotate

liulinboyi:
Hide the line of the Konva rotation handle and keep the rotation function.

Resolve the https://github.com/konvajs/konva/discussions/1688.

```js
var rect = new Konva.Rect({
  x: 100,
  y: 100,
  width: 200,
  height: 200,
  fill: "pink",
  draggable: true
});

var tr = new Konva.Transformer({
  nodes: [rect ],
  anchorSize: 10,
  rotateLinkLineVisiable: false, // add this flag to hide the link line
});
```
The `rotateLinkLineVisiable`'s default value is true, when the value is `false` will be like the follow:
image

### DIFF
--- a/src/shapes/Transformer.ts
+++ b/src/shapes/Transformer.ts
@@ -18,6 +18,7 @@ export interface Box extends IRect {
 export interface TransformerConfig extends ContainerConfig {
   resizeEnabled?: boolean;
   rotateEnabled?: boolean;
+  rotateLinkLineVisiable?: boolean;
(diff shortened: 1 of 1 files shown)


lavrton:
I changed the attribute name a little. Thanks for the PR.

liulinboyi:
> I changed the attribute name a little. Thanks for the PR.

Cool!

PR #1695, opened 2023-12-28
Add support for ctx.roundRect

CPatchane:
According to the [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/roundRect), we should be able to use the `roundRect()` function on the canvas context.

While we can create rounded rectangles by setting `cornerRadius` to `Rect()`, the `roundRect()` function is particularly useful when we want to clip using a rounded rectangle shape with `clipFunc()`.

Since this function appears to be a recent implementation, it might be a good idea to check if it's supported before running it. 

Here I chose to run `rect()` instead so the result would be at least a rectangle but let me know if you think that we should handle this in a different manner.

### DIFF
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -592,12 +592,14 @@ export class Context {
    * @method
    * @name Konva.Context#roundRect
    */
-  roundRect(x: number, y: number, width: number, height: number, radii: number) {
-    if (this._context.roundRect) {
-      this._context.roundRect(x, y, width, height, radii);
-    } else {
-      this._context.rect(x, y, width, height);
-    }
+  roundRect(
+    x: number,
+    y: number,
+    width: number,
+    height: number,
+    radii: number
+  ) {
+    this._context.roundRect(x, y, width, height, radii);
   }
   /**
    * putImageData function.
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -58,6 +58,7 @@ var COMMA = ',',
     'putImageData',
     'quadraticCurveTo',
     'rect',
+    'roundRect',
     'restore',
     'rotate',
     'save',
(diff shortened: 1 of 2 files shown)


lavrton:
I think we can skip the check, as it is already supported by all major browsers.

CPatchane:
It was introduced in Chrome 99 less than two years ago.

I removed the check then 👍

CPatchane:
I just saw that you have a [util function to draw round rectangle](https://github.com/konvajs/konva/blob/a8efcd554ac04531fd3e93d56ecb78ac57518a15/src/Util.ts#L993C1-L993C1) that is used in some `sceneFunc`. Should it be used instead to avoid the browser support issue?

src/Context.ts
```diff
@@ -586,6 +587,14 @@ export class Context {
   rect(x: number, y: number, width: number, height: number) {
     this._context.rect(x, y, width, height);
   }
+  /**
+   * roundRect function.
+   * @method
+   * @name Konva.Context#roundRect
+   */
+  roundRect(x: number, y: number, width: number, height: number, radii: number) {
```
lavrton, line 595: Should type of `radii` by also an array? Like `number | Array<number>`?

lavrton (changes requested):

src/Context.ts
CPatchane: Oh you're right, missed the fact that you can fact an array of corners name 👍

CPatchane: It won't be compatible with [the internal `drawRoundedRectPath`](https://github.com/konvajs/konva/pull/1695#issuecomment-1871972576) function
So let's just use the native API then 👍

lavrton:
Interesting, I just tried to use `ctx.roundRect()` directly in `Konva.Rect` shape when it has corner radius. And it works slower than `drawRoundedRectPath`. I was expecting a different result.

CPatchane:
> Interesting, I just tried to use ctx.roundRect() directly in Konva.Rect shape when it has corner radius. And it works slower than drawRoundedRectPath. I was expecting a different result.

Ouch, it would be better to have at least the same performance, might be an optimisation issue? Or maybe it has more works to do to handle DOMPointInit for radius

PR #1777, opened 2024-06-11
perf: Make index of Node readonly to prevent potential perf problem

luowenxing:
Make index of Node readonly to prevent potential remove child with complexity o(n^2)

Note that **Konva.Container#removeChildren** function comment with these line
> reset parent to prevent many _setChildrenIndices calls

That means without reset the parent, **Konva.Node#remove** would call **_setChildrenIndice** which will take into extra o(n^2) complexity. It works in this case but failed in other case such as call remove node in a certain loop, which offen ocurrs  in increasing draw visible node and decreasing destory invisible node in current viewport.

Thus replace the index property with readonly index is a better way to prevent the unexpected complextiy by frequent operation of a Konva Node.

### DIFF
--- a/src/Container.ts
+++ b/src/Container.ts
@@ -83,7 +83,6 @@ export abstract class Container<
     this.getChildren().forEach((child) => {
       // reset parent to prevent many _setChildrenIndices calls
       child.parent = null;
-      child.index = 0;
       child.remove();
     });
     this.children = [];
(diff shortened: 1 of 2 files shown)


lavrton:
Do you have a specific sample/demo where it gives performance boost?
Currently, it fails tests.

PR #1795, opened 2024-07-18
fix: specify type

kyechan99:
Hello 😄 

Is there a reason for not specifying the type for the following changes?
I changed it to use it more accurately.
Thx.

### Before
```typescript
export declare class Path extends Shape<PathConfig> {
    // ...
    static getPointAtLengthOfDataArray(length: number, dataArray: any): {
        x: any;
        y: any;
    } | null;
    static getPointOnLine(dist: any, P1x: any, P1y: any, P2x: any, P2y: any, fromX?: any, fromY?: any): {
        x: any;
        y: any;
    };
}
```

### After
```typescript
export declare class Path extends Shape<PathConfig> {
    // ...
    static getPointAtLengthOfDataArray(length: number, dataArray: any): Vector2d | null;
    static getPointOnLine(dist: any, P1x: any, P1y: any, P2x: any, P2y: any, fromX?: any, fromY?: any): Vector2d;
}
```

### Before
```typescript
export declare class Text extends Shape<TextConfig> {
    // ...
    measureSize(text: any): {
        actualBoundingBoxAscent: any;
        actualBoundingBoxDescent: any;
        actualBoundingBoxLeft: any;
        actualBoundingBoxRight: any;
        alphabeticBaseline: any;
        emHeightAscent: any;
        emHeightDescent: any;
        fontBoundingBoxAscent: any;
        fontBoundingBoxDescent: any;
        hangingBaseline: any;
        ideographicBaseline: any;
        width: any;
        height: number;
    };
}
```

### After
```typescript
export declare class Text extends Shape<TextConfig> {
    // ...
    measureSize(text: string): {
        actualBoundingBoxAscent: number;
        actualBoundingBoxDescent: number;
        actualBoundingBoxLeft: number;
        actualBoundingBoxRight: number;
        alphabeticBaseline: number;
        emHeightAscent: number;
        emHeightDescent: number;
        fontBoundingBoxAscent: number;
        fontBoundingBoxDescent: number;
        hangingBaseline: number;
        ideographicBaseline: number;
        width: number;
        height: number;
    };
}
```

### DIFF
--- a/src/shapes/Path.ts
+++ b/src/shapes/Path.ts
@@ -2,7 +2,7 @@ import { Factory } from '../Factory';
 import { Shape, ShapeConfig } from '../Shape';
 import { _registerNode } from '../Global';
 
-import { GetSet, PathSegment, Vector2d } from '../types';
+import { GetSet, PathSegment } from '../types';
 import {
   getCubicArcLength,
   getQuadraticArcLength,
@@ -235,10 +235,7 @@ export class Path extends Shape<PathConfig> {
     return pathLength;
   }
 
-  static getPointAtLengthOfDataArray(
-    length: number,
-    dataArray
-  ): Vector2d | null {
+  static getPointAtLengthOfDataArray(length: number, dataArray) {
     var point,
       i = 0,
       ii = dataArray.length;
@@ -322,7 +319,7 @@ export class Path extends Shape<PathConfig> {
     return null;
   }
 
-  static getPointOnLine(dist, P1x, P1y, P2x, P2y, fromX?, fromY?): Vector2d {
+  static getPointOnLine(dist, P1x, P1y, P2x, P2y, fromX?, fromY?) {
     fromX = fromX ?? P1x;
     fromY = fromY ?? P1y;
 
@@ -357,17 +354,7 @@ export class Path extends Shape<PathConfig> {
     return { x: ix + adjustedRun, y: iy + adjustedRise };
   }
 
-  static getPointOnCubicBezier(
-    pct,
-    P1x,
-    P1y,
-    P2x,
-    P2y,
-    P3x,
-    P3y,
-    P4x,
-    P4y
-  ): Vector2d {
+  static getPointOnCubicBezier(pct, P1x, P1y, P2x, P2y, P3x, P3y, P4x, P4y) {
     function CB1(t) {
       return t * t * t;
     }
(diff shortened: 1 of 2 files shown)


lavrton:
Thanks for the PR request. The text changes look good.
For path changes, I prefer not to specify type of return value, because it should be automatically detected by typescript. So if it is not correct (like x is any, instead of number) need to look into internals and fix types inside. Can you do that?

kyechan99:
Understand.
Reverted Path changes.

lavrton:
Well, it may be still good to have some changes, like to check why x and y are `any` while they should be `number`, do you want to do that?

kyechan99:
OK.
How about this?
It can find out that `x` and `y` are `number`s without specifying a return type.
Tell me if it's different from what you think.

## Exampe
```typescript
  static getPointOnCubicBezier(pct, P1x, P1y, P2x, P2y, P3x, P3y, P4x, P4y) {
    function CB1(t: number) {   // <- argument type
      return t * t * t;
    }
    function CB2(t: number) {   // <- argument type
      return 3 * t * t * (1 - t);
    }
    function CB3(t: number) {   // <- argument type
      return 3 * t * (1 - t) * (1 - t);
    }
    function CB4(t: number) {   // <- argument type
      return (1 - t) * (1 - t) * (1 - t);
    }
    var x = P4x * CB1(pct) + P3x * CB2(pct) + P2x * CB3(pct) + P1x * CB4(pct);
    var y = P4y * CB1(pct) + P3y * CB2(pct) + P2y * CB3(pct) + P1y * CB4(pct);

    return {
      x: x,  // number
      y: y,  // number
    };
  }
```

### Path.d.ts
```typescript
Path.getPointOnCubicBezier(pct: any, P1x: any, P1y: any, P2x: any, P2y: any, P3x: any, P3y: any, P4x: any, P4y: any): {
    x: number;
    y: number;
}
```

### To use
```typescript
Path.getPointOnCubicBezier('0', '0', '0', '0', '0', '0', '0', '0', '0');
Path.getPointOnCubicBezier(0, 0, 0, 0, 0, 0, 0, 0, 0);
```

lavrton:
Well, as I can see, the type of `getPointOnCubicBezier` is already `{ x: number;    y: number; }`. So we don't need to do much here with the return type. But I don't think the arguments are correct here. Currently, they have no types. So they are `any`, but I think only numbers should be allowed there.

As I can see `getPointOnLine` has not 100% corect return, Also `getPointAtLength`, `getPointAtLengthOfDataArray`

kyechan99:
Yeah, youre right `getPointOnCubicBezier` doesn't need any modification.

To modify the three functions (return `{ x: any; y: any; }`)
I can only think of two ways to solve it.
1. Specifying the type of argument
2. Using `Number` or `parseInt`.
  
If there is a better way, please let me know.


```typescript
static getPointOnLine(
    dist: number,
    P1x: number,
    P1y: number,
    P2x: number,
    P2y: number,
    fromX?: number,
    fromY?: number
  ) { ... } 
```

lavrton:
> Specifying the type of argument

This is the way.

Also, some functions may need additional types inside their body.

lavrton:
Thanks

PR #1806, opened 2024-08-19
@quarksb refactor: change Filter input and output

### LINKED ISSUE #1804: filters is very slow
the api getImageData is very slow beacause it need gpu convert data to cpu, why not just set the filterContext as the input of filters? which could get filter canvas easily

quarksb:
Resolves #1804 
getImageData and put imageData is very slow, so I change imageData to Canvas, if you want use imageData, you must use getImageData by yourself in Filter

### DIFF
--- a/src/Node.ts
+++ b/src/Node.ts
@@ -15,7 +15,8 @@ import { Context } from './Context';
 import { Shape } from './Shape';
 import { Layer } from './Layer';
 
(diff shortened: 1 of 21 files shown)


lavrton:
Hello, thanks for your pull request. I will also review another one that your submitted.

For that one, I can say that I can't accept it because it is a very breaking change. Some devs may have custom filters, so switching from imageData to canvas will break the code.

PR #1811, opened 2024-08-25
feat: improved types

psychedelicious:
Improved types, mostly in `Node.ts`.

Notes:
- There should no code changes, only typing changes.
- I've explicitly typed the return value of many functions that previously had their return types inferred. This provides insurance against unintentional changes to return types.
- I wrote some docstrings for the export config methods. Please carefully review these for accuracy.

Followups:
- I'm not sure how the API docs are created. I assume generated from the JSDoc annotations? If so, I can work on those.

### DIFF
--- a/src/Node.ts
+++ b/src/Node.ts
@@ -3,7 +3,18 @@ import { Factory } from './Factory';
 import { SceneCanvas, HitCanvas, Canvas } from './Canvas';
 import { Konva } from './Global';
 import { Container } from './Container';
-import { GetSet, Vector2d, IRect } from './types';
+import {
+  GetSet,
+  Vector2d,
+  IRect,
+  ToCanvasConfig,
+  ToDataURLConfig,
+  ToImageConfig,
+  ToBlobConfig,
+  Size,
+  CacheConfig,
+  GetClientRectConfig,
+} from './types';
 import { DD } from './DragAndDrop';
 import {
   getNumberValidator,
@@ -172,11 +183,11 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
     // all change event listeners are attached to the prototype
   }
 
-  hasChildren() {
+  hasChildren(): boolean {
     return false;
   }
 
-  _clearCache(attr?: string) {
+  _clearCache(attr?: string): void {
     // if we want to clear transform cache
     // we don't really need to remove it from the cache
     // but instead mark as "dirty"
@@ -300,17 +311,7 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
    *   drawBorder: true
    * });
    */
-  cache(config?: {
-    x?: number;
-    y?: number;
-    width?: number;
-    height?: number;
-    drawBorder?: boolean;
-    offset?: number;
-    pixelRatio?: number;
-    imageSmoothingEnabled?: boolean;
-    hitCanvasPixelRatio?: number;
-  }) {
+  cache(config?: CacheConfig): typeof this | undefined {
     var conf = config || {};
     var rect = {} as IRect;
 
@@ -445,7 +446,7 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
    * @name Konva.Node#isCached
    * @returns {Boolean}
    */
-  isCached() {
+  isCached(): boolean {
     return this._cache.has(CANVAS);
   }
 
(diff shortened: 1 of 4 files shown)


src/Node.ts
```diff
@@ -915,8 +915,8 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
    * @name Konva.Node#getAttrs
    * @returns {Object}
    */
-  getAttrs() {
-    return (this.attrs || {}) as Config & Record<string, any>;
+  getAttrs(): Partial<Config> {
```
psychedelicious, line 918: I'd like to type `this.attrs` as `Partial<Config>`, but this will require some minor code changes, so I've left it out of this PR.

src/types.ts
```diff
@@ -82,3 +84,148 @@ export interface RGB {
 export interface RGBA extends RGB {
   a: number;
 }
+
+export interface Size {
```
psychedelicious, line 88: Maybe a better name is `Dimensions`?

src/Node.ts
psychedelicious, line 1: I'd like to improve the types of `this._cache`. Currently it is `Map<string, any>`.

I had a couple ideas to improve this, but I'm not sure of the intended use of the cache. Is it intended to be a generic cache with arbitrary keys, or can we be strict about what keys are used for it?

If we can be strict, I was thinking about a simple wrapper:
```ts
const CacheKey = {
  ABSOLUTE_OPACITY: 'absoluteOpacity',
  ALL_LISTENERS: 'allEventListeners',
  ABSOLUTE_TRANSFORM: 'absoluteTransform',
  ... // the rest of the keys
} as const;

type CacheKeyMap = {
  [CacheKey.ABSOLUTE_OPACITY]: number;
  [CacheKey.ALL_LISTENERS]: unknown; // not sure what this should be
  [CacheKey.ABSOLUTE_TRANSFORM]: Transform;
  ... // the rest of the keys & their types
};

class CacheMap extends Map {
  get<T extends keyof CacheKeyMap>(key: T): CacheKeyMap[T] | undefined {
    return super.get(key);
  }
  set<T extends keyof CacheKeyMap>(key: T, value: CacheKeyMap[T]): this {
    return super.set(key, value);
  }
  delete<T extends keyof CacheKeyMap>(key: T): boolean {
    return super.delete(key);
  }
}
```

This merits discussion - and would require code changes - so I've left it out of this PR.

src/Node.ts
lavrton: The issue with this type, as it limits to config only. But it is possible to write ANY custom data into attributes.

src/Node.ts
```diff
@@ -172,11 +183,11 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
     // all change event listeners are attached to the prototype
   }
 
-  hasChildren() {
+  hasChildren(): boolean {
```
lavrton, line 186: Personally, I don't like having return types everywhere. I prefer less code as it is more clean. 

> I've explicitly typed the return value of many functions that previously had their return types inferred. This provides insurance against unintentional changes to return types.

It is a good point. But still I prefer not to write type. Probably is is a good idea to write them on complex functions with many returns. But on small one, with one return, most likely we don't really need it. Especially if it is just one direct value return.

src/Node.ts
```diff
@@ -930,7 +930,7 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
    *   fill: 'red'
    * });
    */
-  setAttrs(config: any) {
+  setAttrs(config?: Partial<Config>) {
```
lavrton, line 933: We can set any data in attrs. Not just what is inside `Config`.

src/Node.ts
```diff
@@ -2002,16 +1999,7 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
...
-    x?: number;
-    y?: number;
-    width?: number;
-    height?: number;
-    pixelRatio?: number;
-    mimeType?: string;
-    quality?: number;
-    callback?: (str: string) => void;
-  }) {
+  toDataURL(config?: ToDataURLConfig): string {
```
lavrton, line 2002: I would prefer to keep this type here directly in the code, so it is not defined somewhere else in the code base. I don't think we reuse this type anywhere.

src/Node.ts
lavrton: I don't think we need additional code just for types. Ok, if it is types only, but not if it is a real class.
_cache is not exposed to public API, so as it is very well tested, I don't think we need to care much about its types.

lavrton (changes requested):

PR #1812, opened 2024-08-25
feat: Factory types

psychedelicious:
Add types to the Factory helpers & validators.

Notes:
- I intended for this to only include changes to types, but it uncovered some very minor code issues. See PR comments.

### DIFF
--- a/src/Factory.ts
+++ b/src/Factory.ts
@@ -1,17 +1,75 @@
 import { Node } from './Node';
+import { GetSet } from './types';
 import { Util } from './Util';
 import { getComponentValidator } from './Validators';
 
-var GET = 'get',
-  SET = 'set';
+const GET = 'get';
+const SET = 'set';
+
+/**
+ * Enforces that a type is a string.
+ */
+type EnforceString<T> = T extends string ? T : never;
+
+/**
+ * Represents a class.
+ */
+type Constructor = abstract new (...args: any) => any;
+
+/**
+ * An attribute of an instance of the provided class. Attributes names be strings.
+ */
+type Attr<T extends Constructor> = EnforceString<keyof InstanceType<T>>;
+
+/**
+ * A function that is called after a setter is called.
+ */
+type AfterFunc<T extends Constructor> = (this: InstanceType<T>) => void;
+
+/**
+ * Extracts the type of a GetSet.
+ */
+type ExtractGetSet<T> = T extends GetSet<infer U, any> ? U : never;
+
+/**
+ * Extracts the type of a GetSet class attribute.
+ */
+type Value<T extends Constructor, U extends Attr<T>> = ExtractGetSet<
+  InstanceType<T>[U]
+>;
+
+/**
+ * A function that validates a value.
+ */
+type ValidatorFunc<T> = (val: ExtractGetSet<T>, attr: string) => T;
+
+/**
+ * Extracts the "components" (keys) of a GetSet value. The value must be an object.
+ */
+type ExtractComponents<T extends Constructor, U extends Attr<T>> = Value<
+  T,
+  U
+> extends Record<string, any>
+  ? EnforceString<keyof Value<T, U>>[]
+  : never;
 
 export const Factory = {
-  addGetterSetter(constructor, attr, def?, validator?, after?) {
+  addGetterSetter<T extends Constructor, U extends Attr<T>>(
+    constructor: T,
+    attr: U,
+    def?: Value<T, U>,
+    validator?: ValidatorFunc<Value<T, U>>,
+    after?: AfterFunc<T>
+  ): void {
     Factory.addGetter(constructor, attr, def);
     Factory.addSetter(constructor, attr, validator, after);
     Factory.addOverloadedGetterSetter(constructor, attr);
   },
-  addGetter(constructor, attr, def?) {
+  addGetter<T extends Constructor, U extends Attr<T>>(
+    constructor: T,
+    attr: U,
+    def?: Value<T, U>
+  ) {
     var method = GET + Util._capitalize(attr);
 
     constructor.prototype[method] =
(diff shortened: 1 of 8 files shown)


src/Node.ts
```diff
@@ -2660,7 +2660,7 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
   rotation: GetSet<number, this>;
   zIndex: GetSet<number, this>;
 
-  scale: GetSet<Vector2d | undefined, this>;
+  scale: GetSet<Vector2d, this>;
```
psychedelicious, line 2663: Either this `GetSet` type is incorrect for the `NodeConfig.scale` type is incorrect. I assume it's this one.

src/Node.ts
```diff
@@ -3102,7 +3102,7 @@ addGetterSetter(Node, 'offsetY', 0, getNumberValidator());
  * node.offsetY(3);
  */
 
-addGetterSetter(Node, 'dragDistance', null, getNumberValidator());
+addGetterSetter(Node, 'dragDistance', undefined, getNumberValidator());
```
psychedelicious, line 3105: This is a subtle API change.

Perhaps we should widen the type of `dragDistance` in `NodeConfig` to `dragDistance?: number | null;`?

src/Node.ts
```diff
@@ -3193,7 +3193,7 @@ addGetterSetter(Node, 'listening', true, getBooleanValidator());
 
 addGetterSetter(Node, 'preventDefault', true, getBooleanValidator());
 
-addGetterSetter(Node, 'filters', null, function (this: Node, val) {
+addGetterSetter(Node, 'filters', undefined, function (this: Node, val) {
```
psychedelicious, line 3196: This is a subtle API change.

Perhaps we should widen the type of `filters` in `NodeConfig` to `filters?: Array<Filter> | null`?

src/Shape.ts
```diff
@@ -843,6 +843,11 @@ export class Shape<
   strokeLinearGradientStartPoint: GetSet<Vector2d, this>;
   strokeLinearGradientEndPoint: GetSet<Vector2d, this>;
   strokeLinearGradientColorStops: GetSet<Array<number | string>, this>;
+  strokeLinearGradientStartPointX: GetSet<number, this>;
+  strokeLinearGradientStartPointY: GetSet<number, this>;
+  strokeLinearGradientEndPointX: GetSet<number, this>;
+  strokeLinearGradientEndPointY: GetSet<number, this>;
+  fillRule: GetSet<CanvasFillRule, this>;
```
psychedelicious, line 850: These annotations were missing. I don't think this changes anything because the Factory adds the `GetSet`s anyways.

src/Validators.ts
psychedelicious, line 1: I _think_ I have understood the intention with these validators - they should create warnings but otherwise let the value through, even if it is invalid.

src/filters/Emboss.ts
```diff
@@ -174,7 +174,7 @@ Factory.addGetterSetter(
   Node,
   'embossDirection',
   'top-left',
-  null,
+  undefined,
```
psychedelicious, line 177: Another subtle API change.

src/filters/Emboss.ts
```diff
@@ -190,7 +190,7 @@ Factory.addGetterSetter(
   Node,
   'embossBlend',
   false,
-  null,
+  undefined,
```
psychedelicious, line 193: Another subtle API change.

src/shapes/Label.ts
```diff
@@ -317,7 +317,10 @@ export class Tag extends Shape<TagConfig> {
     };
   }
 
-  pointerDirection: GetSet<'left' | 'top' | 'right' | 'bottom', this>;
+  pointerDirection: GetSet<
+    'left' | 'top' | 'right' | 'bottom' | typeof NONE,
+    this
+  >;
```
psychedelicious, line 323: The factory sets the initial value to `NONE`.

src/shapes/Transformer.ts
```diff
@@ -1745,8 +1745,6 @@ Factory.addGetterSetter(Transformer, 'ignoreStroke', false);
  */
 Factory.addGetterSetter(Transformer, 'padding', 0, getNumberValidator());
 
-Factory.addGetterSetter(Transformer, 'node');
-
```
psychedelicious, line 1749: I think this was added accidentally - `Transformer` never references `this.node` as far as I can see...

src/Factory.ts
```diff
@@ -1,17 +1,75 @@
...
+  U
+> extends Record<string, any>
+  ? EnforceString<keyof Value<T, U>>[]
+  : never;
 
 export const Factory = {
-  addGetterSetter(constructor, attr, def?, validator?, after?) {
+  addGetterSetter<T extends Constructor, U extends Attr<T>>(
+    constructor: T,
+    attr: U,
```
psychedelicious, line 59: We can make this a bit safer with a type to constrain `attr` to be a `GetSet`.

```ts
type EnforceGetSet<
  T extends Constructor,
  U extends Attr<T>
> = InstanceType<T>[U] extends GetSet<any, any> ? U : never;
```

Then, type `attr` as `attr: EnforceGetSet<T, U>`.

However, this causes a problem in `Canvas.ts` at https://github.com/psychedelicious/konva/blob/9c41a6eed1bba46e80ce282903ac86bb11493e84/src/Canvas.ts#L171

psychedelicious:
For this PR, I was just kinda having fun exploring typescript to see if I could get autocomplete working for the Factory methods. Because it is an internal API, it's not a big deal if it has perfect type safety or not - I'm not dead-set on this PR being accepted.

src/Node.ts
lavrton: Looks like it is not correct, I don't think `scale` can return undefined.

src/Factory.ts
```diff
@@ -79,7 +149,7 @@ export const Factory = {
         key;
 
       if (validator) {
-        val = validator.call(this, val);
+        val = validator.call(this, val, attr);
```
lavrton, line 152: What does this change do?

src/Node.ts
lavrton: The drag distance must be always number. So `undefined` should fallback to default value. I am not really sure what it was `null` here.

src/shapes/Transformer.ts
lavrton: This is an old (deprecated) API. Should be safe to remove.

src/shapes/TextPath.ts
```diff
@@ -551,7 +551,7 @@ Factory.addGetterSetter(TextPath, 'text', EMPTY_STRING);
  * // underline text
  * shape.textDecoration('underline');
  */
-Factory.addGetterSetter(TextPath, 'textDecoration', null);
+Factory.addGetterSetter(TextPath, 'textDecoration', undefined);
```
lavrton, line 554: This probably should be an empty string as in Text class.

PR #1814, opened 2024-08-31
Fix race condition when unregister an event inside handlers

### LINKED ISSUE #1813: Race condition when calling `off` in event handlers
### To reproduce

1. Go to https://jsbin.com/sosumon/edit?html,js,output
2. Click the stage

### Expected

On the first click, it should log both `"handler1"` and `"handler2"`.

### Actual behavior

On the first click, only `"handler1"` is logged.

luin:
Closes #1813

# Test Plan

Make sure the issue mentioned in #1813 is fixed


lavrton:
Tests are not passed. And I am not sure yet how to resolve failed test.

PR #1827, opened 2024-10-04
tsconfig ilb to es2019

tbo47:
to fix this errors:

```bash
src/shapes/Text.ts:565:27 - error TS2550: Property 'trimRight' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2019' or later.

565             match = match.trimRight();
                              ~~~~~~~~~

src/shapes/Text.ts:582:25 - error TS2550: Property 'trimLeft' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2019' or later.

582             line = line.trimLeft();
                            ~~~~~~~~
```

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,7 +5,7 @@
     "target": "ES2018",
     // "sourceMap": true,
     "noEmitOnError": true,
-    "lib": ["ES2015", "dom"],
+    "lib": ["ES2019", "dom"],
     "moduleResolution": "node",
     "declaration": true,
     "removeComments": false,


lavrton:
How do you see that error?

tbo47:
```bash
➜  konva git:(master) tsc -v
Version 5.6.2
➜  konva git:(master) tsc

...

src/shapes/Text.ts:565:27 - error TS2550: Property 'trimRight' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2019' or later.

565             match = match.trimRight();
                              ~~~~~~~~~

src/shapes/Text.ts:582:25 - error TS2550: Property 'trimLeft' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2019' or later.

582             line = line.trimLeft();
                            ~~~~~~~~
```

lavrton:
Hmm... Not sure why I don't see anything...

PR #1828, opened 2024-10-04
defensive check for error in browser

tbo47:
I have an error when using in the browser (angular 18). It's probably because the Global object is not defined in the browser.

With this defensive check, everything works fine.

### DIFF
--- a/src/Global.ts
+++ b/src/Global.ts
@@ -154,7 +154,7 @@ export const Konva = {
    * @memberof Konva
    */
   isDragging() {
-    return Konva['DD'].isDragging;
+    return Konva['DD']?.isDragging;
   },
   isTransforming() {
     return Konva['Transformer']?.isTransforming();


lavrton:
Would it be possible for you to make a demo where the issue is reproducible? It looks like some more profound issue. Drag&drop module must exist.

tbo47:
```
core.mjs:7192 ERROR TypeError: Cannot read properties of undefined (reading 'isDragging')
    at Object.isDragging (Global.js:42:28)
    at Stage._pointermove (Stage.js:417:19)
    at content.addEventListener.passive (Stage.js:305:32)
    at _ZoneDelegate.invokeTask (zone.js:402:33)
    at core.mjs:6714:49
    at AsyncStackTaggingZoneSpec.onInvokeTask (core.mjs:6714:30)
    at _ZoneDelegate.invokeTask (zone.js:401:38)
    at Object.onInvokeTask (core.mjs:7034:25)
    at _ZoneDelegate.invokeTask (zone.js:401:38)
    at ZoneImpl.runTask (zone.js:159:47)
```
It's happening if I only use this kind of import `import { Layer } from 'konva/lib/Layer'`. It's not happening when using `import * as konva from 'konva'`.

But I don't want to use the `*` notation as it prevent tree shaking with webpack.

lavrton:
You have to import core first according to docs: https://github.com/konvajs/konva?tab=readme-ov-file#3-minimal-bundle

PR #1847, opened 2024-10-29
Adjust dblclick trigger conditions

Q-hj:
Undouble-clicking the right or middle mouse button triggers a dblclick event.


lavrton:
Can you make a test?

Issue #1871, opened 2025-01-06
Transformer doesn't change when the angle of Arc changes


tonyzhou1890:
_attrsAffectingSize of Arc has 'innerRadius' and 'outerRadius', without angle and clockwise. Is this for some reason?

Without angle:

![without angle](https://github.com/user-attachments/assets/1df9f2d4-515e-422c-b14c-128edffa4975)

with angle:

![with angle](https://github.com/user-attachments/assets/0b04fcec-aeec-4e87-8330-0fd977b1c0d2)

lavrton:
That is a bug. Do you want to make a Pull Request to fix it?

tonyzhou1890:
> That is a bug. Do you want to make a Pull Request to fix it?

"Sorry, I didn't notice your reply. I have submitted a PR."

lavrton:
Will release it later.

PR #1888, opened 2025-02-17
fix: Transformer doesn't change when the angle of Arc changes #1871

tonyzhou1890:
"angle" and "clockwise" will affect size both, so I added them to Arc.prototype._attrsAffectingSize.


lavrton:
Thanks

PR #1920, opened 2025-04-11
add cloud shape

tbo47:
https://jsfiddle.net/zd4u7wy6/

image


lavrton:
Hello. Thanks for the pull request. I am going to close it, as I think that shape is too rare use case and I would like to keep the library minimal where possible. 

May be good idea to have a separate collection or custom shapes or just a tutorial/doc on the website.

tbo47:
I put the code here but it could be moved anywhere https://github.com/tbo47/konva-cloud

PR #1950, opened 2025-07-23
Fixed corner radius for Konva.Rect negative width/height

Caden-Hornyak:
This PR fixes a bug where negative width or height values are allowed in Konva.Rect, but specifying a cornerRadius breaks rendering. This occurs because cornerRadius is internally capped by half of width and height, which are negative in this case, causing invalid results.

Additionally, the previous PR for Konva.RegularPolygon has been updated to ensure consistent handling of negative dimensions across both shapes.


lavrton:
Thanks a lot.

PR #1960, opened 2025-08-19
check pull requests for `prettier` formatting

nathan-muir:
In my other PR (#1959) I noticed there were some unrelated changes, because of formatting (trailing whitespaces etc).

If it helps, I've prepared this PR to:

 - Add a github workflow to check PRs have applied `prettier` (via `npm run fmt`)
 - Fixes some HTML errors that prevented prettier from running
 - Applies prettier to all files

This does add a bit of code churn because of the addition of trailing commas, which could be disabled in prettier -- https://prettier.io/docs/options#trailing-commas

If you're concerned about this affecting `git blame`, after merging [we can add a `.git-blame-ignore-revs` with the commit hash](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view)


lavrton:
Helpful. But I think we need disable trailing commas

PR #1981, opened 2025-10-07
add ts-ignore to allow tsc to work

tbo47:
For `tsc --removeComments` to run without problem.

I notice that tsconfig is configured with "target": "ES2018". Does it mean there is a esm package of konva?

I maintain this build but would prefer not to have to https://www.npmjs.com/package/konva-es


lavrton:
Version 10 is ES only.

PR #2013, opened 2025-12-25
perf: The same node type shares the same eventListeners to improve perf of construct node

luowenxing:
The **_getProtoListeners** function will copy the **eventListeners** from class prototype to Node instance itself, which means if we create 1000 nodes, it will execute 1000 times copy from prototype to node instances.

However， the prototype of nodes should be immutable, and the caller of **_getProtoListeners** will ensure the **this** pointer to the instance by **call**
image

Thus we don't need to copy **eventListeners** to node instance, instead we store **eventsListeners** to a static member protoListenerMap. It will speed up the performance of create konva node.

image

image

In the case below, the extra 4ms time cost of **_getProtoListeners** will no longger needed. With about 10% perf improvement.

### DIFF
--- a/src/Node.ts
+++ b/src/Node.ts
@@ -2503,31 +2503,24 @@ export abstract class Node<Config extends NodeConfig = NodeConfig> {
   static protoListenerMap = new Map<string, any>();
 
   _getProtoListeners(eventType) {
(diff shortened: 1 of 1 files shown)


lavrton:
Thanks.

PR #2035, opened 2026-03-05
refactor: tsconfig add noImplicitOverride, improve code readability

iaosee:


iaosee:
Set `"noImplicitOverride": true` in tsconfig.json to require the override keyword on any class member that overrides a base class method. This prevents a subclass method from accidentally stopping overriding a base method due to renaming or refactoring in the base class. 

https://www.typescriptlang.org/tsconfig/#noImplicitOverride

This also improves code readability by clearly indicating that the method overrides a base class method.

PR #2055, opened 2026-07-14
fix(Shape): respect imageSmoothingEnabled for fillPatternImage

### LINKED ISSUE #1958: Konva.Shape fill image does not respect Layer or Shape image smoothing setting
**The scenario:**
This one's simple. I'd like to create a repeated checkerboard pattern (for transparency background). The best way to do it in Konva seems to be using a Shape's repeating fill pattern with a simple 2x2 checkerboard image (and scaled using scaleX/Y).

**The problem:**
However, since Konva creates a dummy context to generate the fill pattern, it does not respect the image smoothing setting of the original Shape node or the parent Layer, and there seems to be no way to change that via config. The generate checkerboard fill pattern therefore looks very blurry due to bilinear interpolation scaled up from a 2x2 image when it should be a nearest-pixel interpolation.

**The solution:**
1. Simply use the Shape node's image smoothing setting when creating the dummy context.
2. Allow a separate fill-image-smoothing property that's used with the dummy context.

Solution 2 should be non-breaking.

Can submit a PR if needed.

xianjianlf2:
## What

Makes `fillPatternImage` respect `imageSmoothingEnabled`. Previously a pattern fill always rendered smoothed, ignoring the shape/layer `imageSmoothingEnabled` setting.

## Why

Two paths were dropping the flag:

1. **node-canvas** ignores `imageSmoothingEnabled` for pattern fills entirely and uses a non-standard `patternQuality` property instead, so the plain draw path never honored the flag on the node backend.
2. The **buffer-canvas path** (fill + stroke + opacity, or shadow) draws into the shared `stage.bufferCanvas` context, which never inherited the destination context's `imageSmoothingEnabled` — so patterns drawn via the buffer were always smoothed, including in the browser.

## How (2 source edits, ~16 lines)

- `src/Context.ts` `SceneContext._fillPattern`: mirror the context's `imageSmoothingEnabled` onto `patternQuality` (`'good'`/`'nearest'`), guarded by `'patternQuality' in context` so it is a no-op in browsers.
- `src/Shape.ts` `drawScene` buffer branch: propagate `bufferContext.imageSmoothingEnabled = context.imageSmoothingEnabled` inside the existing save/restore (no state leak).

## Testing

- Added `fillPatternImage should respect imageSmoothingEnabled` in `test/unit/Shape-test.ts`; confirmed it fails before (blended boundary pixels) and passes after (pure 0/255 with smoothing off).
- Verified the buffer path end-to-end (fill+stroke+opacity): 0 blended pixels with smoothing off, present with smoothing on.
- Full node suite: 804 passing, 0 failing. `tsc --noEmit` clean.

Closes #1958

### DIFF
--- a/src/Context.ts
+++ b/src/Context.ts
@@ -842,6 +842,18 @@ export class SceneContext extends Context {
     shape._fillFunc(this);
   }
   _fillPattern(shape: Shape) {
+    // `imageSmoothingEnabled` (set on the layer/context) controls pattern
+    // smoothing in the browser. node-canvas ignores it for pattern fills and
+    // instead exposes a non-standard `patternQuality` property, so mirror the
+    // current smoothing setting onto it to keep both backends consistent.
+    const context = this._context as CanvasRenderingContext2D & {
+      patternQuality?: string;
+    };
+    if ('patternQuality' in context) {
+      context.patternQuality = context.imageSmoothingEnabled
+        ? 'good'
+        : 'nearest';
+    }
     this.setAttr('fillStyle', shape._getFillPattern());
     shape._fillFunc(this);
   }
--- a/src/Shape.ts
+++ b/src/Shape.ts
@@ -656,6 +656,10 @@ export class Shape<
         bufferContext.clear();
       }
       bufferContext.save();
+      // the buffer canvas is a separate context that does not inherit the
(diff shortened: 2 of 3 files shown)


lavrton:
Thanks

